PR #58, opened 2024-09-02
feat: Update State for All Contracts

joemonem:
Added the ability to update `state` for each contract.


SlayerAnsh (changes requested):

SlayerAnsh (changes requested):

contracts/hub/router/src/execute.rs
```diff
@@ -431,3 +432,96 @@ pub fn execute_native_receive_callback(
...
+    };
+
+    let verified_admin = if let Some(ref admin) = admin {
+        Some(deps.api.addr_validate(&admin.as_str())?)
+    } else {
+        None
+    };
+
+    let msg = CosmosMsg::Wasm(WasmMsg::Execute {
+        contract_addr: state.virtual_balance_address.unwrap().to_string(),
```
SlayerAnsh: Nitpick: Check for this to be valid is done above, would be better to unwrap it after its been verified that it exists. Unwrap is kind of dangerous for execution in ibc messages (we do have wrappers so it might not affect but just to be on safe side we should avoid unwrap)

contracts/hub/vlp/src/execute.rs
```diff
@@ -611,3 +611,58 @@ pub fn update_fee(
         .add_event(simple_event())
         .add_attribute("action", "update_fee"))
 }
+
+pub fn update_state(
+    deps: DepsMut,
+    info: MessageInfo,
+    pair: Option<Pair>,
```
SlayerAnsh: We can never update pair and total lp tokens

contracts/liquidity/factory/src/execute.rs
```diff
@@ -807,6 +807,139 @@ pub fn execute_withdraw_virtual_balance(
...
 }
 
+pub fn execute_update_state(
+    deps: DepsMut,
+    info: MessageInfo,
+    router_contract: Option<String>,
+    admin: Option<String>,
+    escrow_code_id: Option<u64>,
+    cw20_code_id: Option<u64>,
+    chain_uid: Option<ChainUid>,
```
SlayerAnsh: cannot change chain uid

SlayerAnsh (changes requested):

SlayerAnsh (approved):

PR #68, opened 2024-09-30
Internal Security Audit Fixes

SlayerAnsh:
# Pull Request

Multiple security fixes along with using virtual balance tokens.

## Description

- Security Fix for Funds - Security bug related to iter for funds there a single denom could be used for multiple tokens resulting in inconsistent funds in escrow and vsl. Added new implementation `FundManager` that handles funds. Additional checks related to extra funds being sent are also added
- Refactor - Arranged utils, cw20 msgs and some other codes to increase readability and prevent developer bugs
- Refactor + Bug - Added TokenWithAmount and related implementation to prevent assigning liquidity to different funds. It was a bug in provide liquidity where user sent token_1_liquidity was assignined to token_2_liquidity because pair ordering was wrong
- Remove Indirect Update State method because they introduce complexity in tracking sensitive state updates. These will be added when needed because managed contracts only needs to be migratable, rest they probably will never update internal state because of state sync issues
- There was no verification for cross chain addresses, which provides flexibility but can result in unexpected fund distribution due to small typo errors. Now with cross chain addresses verification, all chain uids should be valid. Note - We cannot validate addresses on VSL for cross chains
- Removed reply for voucher mint because they were not used and hide errors due to cosmwasm error bubbling
- Security Update - Pairs can only be created if there is atleast 1 registered denom. Combined it with our existing security where you cannot add new denom to exisiting token, this makes permissionless update easy
- Events for liquidity change added
- Add liquidity added check for funds being transfered to it, similar to swap check
- Feature - Vouchers now can be used for create pool, add liquidity, and swap.

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Breaking change
- [x] Documentation update
- [x] Refactor/Chore

## Related Issue

## Testing

Steps to test:

1. [Step 1]
2. [Step 2]
3. [Step 3]

## Checklist

- [x] Self-review completed

[Add screenshots here]

## Additional Notes

### DIFF
--- a/contracts/hub/router/src/execute.rs
+++ b/contracts/hub/router/src/execute.rs
@@ -9,42 +9,24 @@ use euclid::{
     events::{tx_event, TxType},
     msgs::{
         router::{ExecuteMsg, RegisterFactoryChainType},
-        virtual_balance::{ExecuteBurn, ExecuteMsg as VirtualBalanceExecuteMsg},
+        virtual_balance::ExecuteBurn,
     },
     timeout::get_timeout,
     token::Token,
-    utils::generate_tx,
+    utils::tx::generate_tx,
     virtual_balance::BalanceKey,
 };
 use euclid_ibc::msg::{ChainIbcExecuteMsg, HubIbcExecuteMsg};
 
 use crate::{
     ibc::receive,
+    query::verify_cross_chain_addresses,
     reply::VIRTUAL_BALANCE_BURN_REPLY_ID,
     state::{
         State, CHAIN_UID_TO_CHAIN, CHANNEL_TO_CHAIN_UID, DEREGISTERED_CHAINS, ESCROW_BALANCES,
         STATE,
     },
 };
-
-// Function to update the pool code ID
-pub fn execute_update_vlp_code_id(
-    deps: DepsMut,
-    info: MessageInfo,
-    new_vlp_code_id: u64,
-) -> Result<Response, ContractError> {
-    let mut state = STATE.load(deps.storage)?;
-
-    ensure!(info.sender == state.admin, ContractError::Unauthorized {});
-
-    state.vlp_code_id = new_vlp_code_id;
-
-    STATE.save(deps.storage, &state)?;
-
-    Ok(Response::new()
-        .add_attribute("method", "update_pool_code_id")
-        .add_attribute("new_vlp_code_id", new_vlp_code_id.to_string()))
-}
 
 pub fn execute_update_lock(deps: DepsMut, info: MessageInfo) -> Result<Response, ContractError> {
     let mut state = STATE.load(deps.storage)?;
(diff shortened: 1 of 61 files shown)


contracts/hub/router/src/execute.rs
```diff
@@ -249,6 +231,14 @@ pub fn execute_withdraw_voucher(
     cross_chain_addresses: Vec<CrossChainUserWithLimit>,
     timeout: Option<u64>,
 ) -> Result<Response, ContractError> {
+    verify_cross_chain_addresses(
```
joemonem, line 232: Wouldn't it be more efficient to just let the function take`CrossChainUserWithLimit` and ignore the limit in the function?

SlayerAnsh: Yes, I was also thinking same but the function never uses limit so it might be wrong pattern to send a struct thats not the real struct the function is verifying. In future if there is a function that verifies cross chain address with limit, it might cause a conflict.

contracts/hub/router/src/ibc/receive.rs
```diff
@@ -222,22 +227,53 @@ fn execute_request_pool_creation(
...
             ContractError::new("Cannot use already existing token without registering it first")
         )
     }
 
+    ensure!(
+        one_token_already_exists,
+        ContractError::new("Cannot create pool with two new tokens")
+    );
+
     // If vlp is already there, send execute msg to it to register the pool, else create a new pool with register msg attached to instantiate msg
```
joemonem, line 275: nitkpick: I'd say it would be clearer if we add this line here directly before using it
```rust
let vlp = VLPS.may_load(deps.storage, pair.get_tupple())?;
```

contracts/liquidity/cw20/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "cw20"
-version = "0.1.0"
-authors = ["Anshudhar Kumar Singh <anshudhar2001@gmail.com>"]
+version = "0.2.0"
+authors = ["Anshudhar Kumar Singh <anshudhar2001@gmail.com>", "Joe Momem"]
```
joemonem: "Joe Monem <jmonem@icloud.com>"

contracts/liquidity/escrow/src/execute.rs
```diff
@@ -3,20 +3,22 @@ use cosmwasm_std::{
...
 pub fn execute_add_allowed_denom(
     deps: DepsMut,
     _env: Env,
     info: MessageInfo,
     denom: TokenType,
 ) -> Result<Response, ContractError> {
+    // Vouchers are not escrowed
+    ensure!(
+        !denom.is_voucher(),
+        ContractError::UnsupportedDenomination {}
```
joemonem: Error message could be more specific, like Vouchers Cannot Be Escrowed

packages/euclid/src/msgs/escrow/msg.rs
```diff
@@ -44,6 +46,13 @@ pub enum QueryMsg {
 #[cw_serde]
 pub struct MigrateMsg {}
 
+#[cw_serde]
+pub struct StateResponse {
+    pub token_id: Token,
```
joemonem: Sometimes variables of type `Token` are named `token_id` and other times `token`. Should we decide on a standard? For example in `FactoryCw20HookMsg { Deposit }` it uses `token: Token`

SlayerAnsh: Yes, lets do a cleanup in next updates. I have updated state response to use token

joemonem (approved):

PR #71, opened 2024-11-04
feat: Cross Chain Limits

joemonem:
# Pull Request

## Description
Created a `Limit` enum that contains Less than or equal, equal, and greater or equal. 
Adjusted Reply flow (will remove if unnecessary)
Optional `preferred_denom` during `Withdraw`
Optional `forwarding_message` of type `CosmosMsg` to be sent from Escrow's withdraw 

## Type of Change

- [ ] Bug fix
- [x ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor/Chore

## Related Issue
https://app.clickup.com/t/86eqm5c36

## Testing

No tests were created. 

## Checklist

- [ x] Self-review completed

N/A

## Additional Notes
N/A


contracts/hub/router/src/reply.rs
```diff
@@ -261,32 +266,65 @@ pub fn on_swap_reply(deps: DepsMut, env: Env, msg: Reply) -> Result<Response, Co
...
+                    .add_attribute("asset_out", swap_msg.asset_out.to_string())
+                    .add_attribute("asset_in", swap_msg.asset_in.token.to_string())
+                    .add_attribute("asset_type", swap_msg.asset_in.token_type.get_key())
+                    .add_attribute("amount_in", swap_msg.amount_in), // .set_data(to_json_binary(&ack)?)
+            )
+        }
+    }
+}
+
+pub fn on_release_escrow_reply(
```
SlayerAnsh: This can be removed because then we would need to cover cases for return type. If its swap then swap response, if its remove liquidity or withdraw message then it will be different ack type

contracts/hub/router/src/reply.rs
```diff
@@ -191,6 +193,9 @@ pub fn on_remove_liquidity_reply(
                         sender: remove_liquidity_tx.sender.clone(),
                         token: token.token.clone(),
                         amount: Some(token.amount),
+                        //TODO
+                        preferred_denom: None,
```
SlayerAnsh: Should be fine as `None` for remove liquidity, as the message can become too complex.

contracts/hub/router/src/reply.rs
```diff
@@ -261,32 +266,65 @@ pub fn on_swap_reply(deps: DepsMut, env: Env, msg: Reply) -> Result<Response, Co
                 amount_out: vlp_swap_response.amount_out,
                 tx_id: vlp_swap_response.tx_id,
             };
-            let ack = AcknowledgementMsg::Ok(swap_response.clone());
+            // let ack = AcknowledgementMsg::Ok(swap_response.clone());
 
             // Prepare burn msg
             let release_msg = ExecuteMsg::ReleaseEscrowInternal {
                 sender: swap_msg.sender,
+                preferred_denom: None,
```
SlayerAnsh: This should be set as preferred denom type provided during inital message as its needed for the forwarding

contracts/hub/router/src/execute.rs
```diff
@@ -265,6 +269,8 @@ pub fn execute_release_escrow(
     info: MessageInfo,
     sender: CrossChainUser,
     token: Token,
+    preferred_denom: Option<TokenType>,
```
SlayerAnsh: We can add a check for correct denom using `TOKEN_DENOM` state.

contracts/hub/vlp/src/execute.rs
```diff
@@ -144,7 +144,7 @@ pub fn add_liquidity(
     liquidity: PairWithAmount,
     slippage_tolerance_bps: u64,
     tx_id: String,
-    called_by_register_pool_with_funds: bool,
+    _called_by_register_pool_with_funds: bool,
```
SlayerAnsh: We can remove this

contracts/liquidity/escrow/src/execute.rs
```diff
@@ -257,11 +287,17 @@ pub fn execute_withdraw(
...
-    Ok(Response::new()
+    let mut response = Response::new()
         .add_messages(messages)
         .add_attribute("method", "escrow_withdraw")
         .add_attribute("amount", amount)
         .add_attribute("token", state.token_id.to_string())
-        .add_attribute("recipient", recipient))
+        .add_attribute("recipient", recipient);
+
+    if let Some(forwarding_message) = forwarding_message {
```
SlayerAnsh: Forwarding message will either send msg with funds or a cw20 send msg depending on the fund type. Goal is to send the funds as msg to recipient if forwarding msg is provided else send the funds as normal funds (like how its currently sending the msgs).

contracts/liquidity/escrow/src/execute.rs
```diff
@@ -257,11 +287,17 @@ pub fn execute_withdraw(
...
+        let msg: CosmosMsg = from_json(&forwarding_message)?;
+        response = response.add_message(msg);
```
SlayerAnsh: Forwarding message will also need a reply id which if the message fails, it will not revert the full tx instead send funds to refund address. If refund address is not provided then revert the tx by failing release completely (which should already be handled in router to mint vouchers again)

SlayerAnsh (changes requested):

contracts/liquidity/escrow/src/execute.rs
joemonem: Should the refund address be a separate optional variable like `forwarding_message`?

contracts/liquidity/escrow/src/execute.rs
joemonem: So should we match the deserialization of the CosmosMsg into 
```rust
            CosmosMsg::Bank(_) => {}
            CosmosMsg::Wasm(_) => {}
```

SlayerAnsh: If msg is None then use the normal send flow which we have currently, if its not None then check if asset is Native, if its native then a direct Wasm msg recipient with funds attached. If asset is cw20 then create a cw20 send msg with the amount and then send Wasm msg to cw20 contract.

contracts/liquidity/escrow/src/execute.rs
SlayerAnsh: Yes

SlayerAnsh (approved):
LGTM!

PR #79, opened 2024-12-11
Forwarding Contract and Refund Logic fix in escrow forwarding

SlayerAnsh:
# Pull Request

## Description

- Add Forwarding Contract for astroport
- Update escrow refund logic for forwarding to handle previous states and partial refund based on which forward hop was failed
- Added forwarding message helper in `create_transfer_msg` to wrap msgs properly based on msgs being forwarded

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Breaking change
- [ ] Documentation update
- [ ] Refactor/Chore

## Related Issue

## Testing

Steps to test:

1. [Step 1]
2. [Step 2]
3. [Step 3]

## Checklist

- [ ] Self-review completed

[Add screenshots here]

## Additional Notes


contracts/forwarding/astroport-forwarding/src/reply.rs
```diff
@@ -0,0 +1,44 @@
+use cosmwasm_std::{ensure, DepsMut, Env, Reply, Response, SubMsgResult};
+use euclid::error::ContractError;
+
+use crate::state::FORWARDING_STATE;
+
+pub const ASTRO_SWAP_REPLY_ID: u64 = 1;
+
+pub fn on_astro_swap_reply(deps: DepsMut, env: Env, msg: Reply) -> Result<Response, ContractError> {
+    match msg.result.clone() {
+        SubMsgResult::Err(err) => Err(ContractError::new(&format!(
```
joemonem, line 10: No changes are made to the `FORWARDING_STATE` in case the swap returns an error?

contracts/forwarding/astroport-forwarding/src/execute.rs
```diff
@@ -0,0 +1,122 @@
...
+};
+
+pub fn execute_cw20_receive(
+    deps: &mut DepsMut,
+    env: &Env,
+    info: &MessageInfo,
+    msg: Cw20ReceiveMsg,
+) -> Result<Response, ContractError> {
+    // let state = STATE.load(deps.storage)?;
+    // let sender = msg.sender;
```
joemonem: The original sender won't be used for any potential refunds? If not we can delete those two comments

SlayerAnsh: We are going to fail messages if there is no refund address and error will be handled by escrow refund logic.

Will remove these two lines

contracts/forwarding/astroport-forwarding/src/reply.rs
SlayerAnsh: Yes coz it error and we also error in forwarding contract, funds will automatically transfer back to escrow and it will handle it

joemonem: Yeah i think the funds are handled, but what about the state after an error in the reply? Won't the contract get stuck at this?
```rust
    ensure!(
        FORWARDING_STATE.may_load(deps.storage)?.is_none(),
        ContractError::new("Contract already in forwarding state")
    );
```

joemonem: Also maybe we could make the above check earlier, possible at the beginning of the `ExecuteMsg` entry point?

SlayerAnsh: Yes, this check is probably not needed. Added it for multihop which will be handled by messages instead

SlayerAnsh: Removing it

joemonem (approved):
Just needs linting fix

PR #83, opened 2024-12-30
Fix - forwarding msg vulnerability

SlayerAnsh:
# Pull Request

## Description

Fix Vulnerability where forwarding message can execute any message on behalf of escrow. Now Forwarding messages are always typed as EuclidReceive and Contracts need to implement it to interact with Euclid related forwarding messages

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Breaking change
- [ ] Documentation update
- [ ] Refactor/Chore

## Related Issue

## Testing

Steps to test:

1. [Step 1]
2. [Step 2]
3. [Step 3]

## Checklist

- [x] Self-review completed

[Add screenshots here]

## Additional Notes

### DIFF
--- a/contracts/forwarding/astroport-forwarding/src/contract.rs
+++ b/contracts/forwarding/astroport-forwarding/src/contract.rs
@@ -44,7 +44,7 @@ pub fn execute(
     msg: ExecuteMsg,
 ) -> Result<Response, ContractError> {
     match msg {
-        ExecuteMsg::Swap(msg) => execute_forward(deps.borrow_mut(), &env, &info, msg),
+        ExecuteMsg::EuclidReceive(msg) => execute_forward(deps.borrow_mut(), &env, &info, msg),
         ExecuteMsg::Receive(msg) => execute_cw20_receive(deps.borrow_mut(), &env, &info, msg),
     }
 }
--- a/contracts/forwarding/astroport-forwarding/src/contract.rs
+++ b/contracts/forwarding/astroport-forwarding/src/contract.rs
@@ -2,13 +2,13 @@ use std::borrow::BorrowMut;
 
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
-use cosmwasm_std::{Binary, Deps, DepsMut, Env, MessageInfo, Reply, Response, StdError};
+use cosmwasm_std::{ensure, Binary, Deps, DepsMut, Env, MessageInfo, Reply, Response, StdError};
 
 use cw2::set_contract_version;
-use euclid::error::ContractError;
+use euclid::{error::ContractError, token::TokenType};
 
 use crate::{
-    execute::{execute_cw20_receive, execute_forward},
+    execute::{execute_cw20_receive, receive_euclid_native, swap},
     reply::{on_astro_swap_reply, ASTRO_SWAP_REPLY_ID},
     state::{State, STATE},
 };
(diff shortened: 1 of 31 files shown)


contracts/forwarding/astroport-forwarding/src/contract.rs
```diff
@@ -44,8 +44,22 @@ pub fn execute(
...
 ) -> Result<Response, ContractError> {
     match msg {
-        ExecuteMsg::Swap(msg) => execute_forward(deps.borrow_mut(), &env, &info, msg),
+        ExecuteMsg::EuclidReceive(msg) => {
+            receive_euclid_native(deps.borrow_mut(), &env, &info, msg)
+        }
         ExecuteMsg::Receive(msg) => execute_cw20_receive(deps.borrow_mut(), &env, &info, msg),
+        ExecuteMsg::Swap(swap_msg) => {
+            ensure!(
+                info.funds.len() == 1,
```
joemonem, line 53: We could use the new `validate_n_funds` function here

contracts/liquidity/factory/src/execute.rs
```diff
@@ -835,6 +839,102 @@ pub fn receive_cw20(
...
+                ExecuteMsg::ExecuteSwapRequest(swap_msg),
+            )?;
+            let event = simple_event()
+                .add_attribute("meta", euclid_receive.meta.unwrap_or("no_meta".to_string()));
+            Ok(response.add_event(event))
+        }
+    }
+}
+
+pub fn receive_euclid_cw20(
```
joemonem, line 895: It's still possible to call `Swap` directly from `receive_cw20`, is this intentional?

contracts/liquidity/factory/src/execute.rs
```diff
@@ -835,6 +839,102 @@ pub fn receive_cw20(
                 &mut deps, env, info, sender, asset_in, amount_in, timeout, recipient,
             )
         }
+        FactoryCw20HookMsg::EuclidReceive(euclid_receive) => {
+            receive_euclid_cw20(deps, env, info, sender, cw20_msg.amount, euclid_receive)
+        }
+    }
+}
+
+pub fn receive_euclid_native(
```
joemonem, line 848: Same question, it's still possible to call Swap directly from the `ExecuteMsg`

packages/euclid/src/msgs/hook.rs
```diff
@@ -0,0 +1,25 @@
...
+}
+
+// This is just a helper to properly serialize the above message
+#[cw_serde]
+pub enum EuclidReceiverMsg {
+    EuclidReceive(EuclidReceive),
+}
+
+impl EuclidReceive {
+    pub fn to_cosmos_msg(&self) -> Result<Binary, ContractError> {
```
joemonem: The function name is a bit misleading since it returns a `Binary` and not a `CosmosMsg`

contracts/forwarding/astroport-forwarding/src/contract.rs
SlayerAnsh: Created for this but intentionally not using it here because it will be extra computation to convert array to hash and then use first element

contracts/liquidity/factory/src/execute.rs
SlayerAnsh: Yes, it will be used for cw20 tokens as asset in token.

contracts/liquidity/factory/src/execute.rs
SlayerAnsh: Yes, it will be used in normal flow, Euclid Receive is mainly used by contracts to safely call any euclid compatible contracts. This way, a Binary message cannot call `Send` message in cw20 because it will be wrapped in EuclidReceive {Send} which is not same as Send hence protecting the contract from unauthorized executions.

packages/euclid/src/msgs/hook.rs
SlayerAnsh: Changed to `to_receiver_msg`

joemonem:
Should we bump the `euclid` package version?

PR #102, opened 2025-05-07
Feat: Relayer Contract

SlayerAnsh:
# Pull Request

## Description

Added relayer contract that can verify meta transactions, and also updated tests

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor/Chore

## Checklist

- [x] Self-review completed


packages/relayer/src/msgs.rs
```diff
@@ -0,0 +1,61 @@
+use cosmwasm_schema::{cw_serde, QueryResponses};
+use cosmwasm_std::{Addr, Binary};
+
+#[cw_serde]
+pub struct InstantiateMsg {
+    pub relayer_pubkey: Binary,
+    pub relayer_address: String,
```
joemonem, line 7: Would be useful to add a comment explaining what this refers to. Since it's strange including a relayer address while instantiating a relayer contract

contracts/common/relayer/src/execute.rs
```diff
@@ -0,0 +1,108 @@
...
+
+pub fn execute_execute_meta_transaction(
+    deps: &mut DepsMut,
+    env: &Env,
+    info: &MessageInfo,
+    msg: MetaTransaction,
+) -> Result<Response, ContractError> {
+    // Ensure the nonce is not used
+    ensure!(
+        !NONCES.has(deps.storage, info.sender.to_string()),
```
joemonem: How could a previous nonce be the sender's address? Are we using addresses as nonces? In the integration tests the nonce is a sequence

contracts/common/relayer/src/state.rs
```diff
@@ -0,0 +1,8 @@
+use cosmwasm_std::Uint128;
+use cw_storage_plus::{Item, Map};
+use relayer::msgs::State;
+
+pub const STATE: Item<State> = Item::new("state");
+
+// nonce -> block height when it was relayed
+pub const NONCES: Map<String, Uint128> = Map::new("nonces");
```
joemonem: Noting that in the Solidity version we're mapping to a `bool`. Is there a reason for the difference?

SlayerAnsh: Resolved in solidity contract, we need block height for quick reference to block a tx was realyed incase state is pruned from chain

contracts/common/relayer/src/execute.rs
SlayerAnsh: Bug, resolved it now.

joemonem (approved):

PR #103, opened 2025-06-05
Migrate contract

joemonem:
# Pull Request

## Description
Migration contract that allows cross-chain migration for our Hub contracts

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor/Chore

## Related Issue

## Testing
`migrate.rs` integration test

## Checklist

- [ ] Self-review completed

[Add screenshots here]

## Additional Notes
We need to check if we're missing or we should exclude some states in the queries, also we need to see if modification is needed for the results. For example: the router keeps the code id of the vlp, but that code id will be different on the chain we're migrating to.


joemonem:
No longer needed

PR #106, opened 2025-07-21
refactor: Remove Zero Allowance and Balance from Virtual Balance

joemonem:
# Pull Request

## Description
After burning or transferring, if the new balance or allowance is zero, it's removed from state. 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor/Chore

## Related Issue

## Testing

Steps to test:

1. [Step 1]
2. [Step 2]
3. [Step 3]

## Checklist

- [ ] Self-review completed

[Add screenshots here]

## Additional Notes

### DIFF
--- a/contracts/hub/virtual_balance/src/execute.rs
+++ b/contracts/hub/virtual_balance/src/execute.rs
@@ -1,4 +1,4 @@
-use cosmwasm_std::{ensure, Addr, DepsMut, MessageInfo, Response, Uint128};
+use cosmwasm_std::{ensure, Addr, DepsMut, MessageInfo, Order, Response, Uint128};
 use euclid::{
     chain::ChainUid,
     error::ContractError,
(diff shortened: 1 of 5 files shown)


contracts/hub/virtual_balance/src/execute.rs
```diff
@@ -61,14 +61,20 @@ pub fn execute_burn(
...
 
     let key = msg.balance_key.clone().to_serialized_balance_key();
-
-    let old_balance = BALANCES
-        .may_load(deps.storage, key.clone())?
-        .unwrap_or(Uint128::zero());
+    let old_balance =
+        BALANCES
+            .may_load(deps.storage, key.clone())?
+            .ok_or(ContractError::BalanceNotFound {
```
SlayerAnsh, line 67: We don't want to error if old balance is not found.

SlayerAnsh (changes requested):

contracts/hub/virtual_balance/src/execute.rs
joemonem: What's the reasoning behind this? We were defaulting to 0 if the balance wasn't found, and then subtracting it with a value that's been verified to be greater than 0, guaranteeing an error from the `checked_sub` operation

SlayerAnsh: Yes make sense. My concern was most of the logic in vcoin treat no balance found as zero balance, so if we error during balance getter function it might introduce hard to debug errors in flow.
You are correct, we can safely add zero balance error here.

contracts/hub/virtual_balance/src/execute.rs
```diff
@@ -291,3 +291,42 @@ pub fn execute_approve(
...
+
+pub fn execute_remove_zero_state_values(
+    deps: DepsMut,
+    info: MessageInfo,
+) -> Result<Response, ContractError> {
+    let state = STATE.load(deps.storage)?;
+
+    // Sender should either be router or admin
+    ensure!(
+        state.router == info.sender.to_string() || (state.admin == info.sender),
```
SlayerAnsh, line 303: It should be handled by admin only

SlayerAnsh (changes requested):

SlayerAnsh (approved):

PR #109, opened 2025-08-07
fix: add support for compressed pubkey + test

SlayerAnsh:
# Pull Request

## Description

Added support for compressed pubkey for verify_signature

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor/Chore

## Related Issue

## Testing

Steps to test:

1. [Step 1]
2. [Step 2]
3. [Step 3]

## Checklist

- [ ] Self-review completed

[Add screenshots here]

## Additional Notes


packages/relayer/Cargo.toml
```diff
@@ -25,6 +25,5 @@ euclid = { workspace = true }
 sha2 = "0.10.8"
 
 cw-orch = { workspace = true }
-
-[dev-dependencies]
 k256 = "0.13.4"
+getrandom = { version = "0.2", features = ["js"] }
```
joemonem: Why was this added? Can't see it being used anywhere

packages/relayer/src/verify.rs
```diff
@@ -80,6 +80,12 @@ pub fn msg_to_sign_data(msg: Binary, signer: String) -> MsgSignData {
     MsgSignData::new(vec![msg_sign_data_msg])
 }
 
+pub fn get_k256_pubkey(pubkey: &Binary) -> Result<k256::ecdsa::VerifyingKey, ContractError> {
```
joemonem: Would be nicer if `VerifyingKey` was imported from the top to avoid this long text

packages/relayer/src/verify.rs
```diff
@@ -89,8 +95,15 @@ pub fn verify_signature(
     let message_hash: [u8; 32] = Sha256::digest(message).into();
     let signature_bytes = signature.as_slice();
     let pubkey_bytes = pubkey.as_slice();
+    // If pubkey is compressed (33 bytes), uncompress it
+    let pubkey_bytes = if pubkey_bytes.len() == 33 {
```
joemonem: The comment is good and doesn't make a `const` necessary, but I think it would be nice to have, just in case that value is reused elsewhere

packages/relayer/src/verify.rs
```diff
@@ -121,6 +134,22 @@ mod tests {
         (secret_key, pub_key)
     }
 
+    fn get_signer_key_from_priv_key(priv_key: &str) -> (SigningKey, Binary) {
+        let scalar = NonZeroScalar::from_str(priv_key).unwrap();
```
joemonem, line 125: Is the use of `unwrap()` avoidable?

packages/relayer/src/verify.rs
```diff
@@ -132,6 +161,16 @@ mod tests {
         (Binary::from(signature.to_vec()), public_key_bytes)
     }
 
+    fn sign_messsage_from_priv_key(msg: &str, priv_key: &str) -> (Binary, Binary) {
+        let message_digest = Sha256::new().chain(msg.as_bytes());
+        let (secret_key, public_key_bytes) = get_signer_key_from_priv_key(priv_key);
+        let signature = secret_key
+            .sign_digest_recoverable(message_digest)
+            .unwrap()
```
joemonem, line 156: Is the use of `unwrap()` avoidable?

packages/relayer/Cargo.toml
SlayerAnsh: It was error due to js feature, without it wasm32 build will fail

packages/relayer/src/verify.rs
SlayerAnsh: Its a test file so kept unwrap for linear code

packages/relayer/src/verify.rs
SlayerAnsh: Test file, unwrap is used intentionally to avoid multiple Result wrap

PR #111, opened 2025-08-11
feat: Concentrated VLP

joemonem:
# Pull Request

## Description

[Provide a brief summary of the changes and their purpose]

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor/Chore

## Related Issue

## Testing

Steps to test:

1. [Step 1]
2. [Step 2]
3. [Step 3]

## Checklist

- [ ] Self-review completed

[Add screenshots here]

## Additional Notes


packages/euclid/src/msgs/concentrated_vlp.rs
```diff
@@ -0,0 +1,1058 @@
...
+    pub router: String,
+    pub virtual_balance: String,
+    pub fee: Fee,
+    pub execute: Option<ExecuteMsg>,
+    pub admin: String,
+    // Concentrated VLP
+    /// The pair type
+    pub pair_type: PairType,
+    /// Asset information for the assets in the pool
+    pub asset_infos: Vec<AssetInfo>,
```
SlayerAnsh: We should try to make this structure similar to pool, all the configs will have have a common base structure and any extra configuration is added via PoolType enum like the Amp Factor for stable pool

SlayerAnsh: Or we update this in router to handle based on pool type so we don't have to maintain a common structure. Router will instead create a enum and extend pools

contracts/hub/concentrated_vlp/src/testing/tests.rs
```diff
@@ -0,0 +1,147 @@
...
+            slippage_tolerance: None,
+            auto_stake: None,
+            receiver: None,
+            min_lp_to_receive: None,
+        };
+
+        let router = deps.api.addr_make("router");
+        let info = message_info(&router, &coins(1000, "earth"));
+
+        // Store precissions, not sure when or where to do this in production
```
SlayerAnsh: Can you elaborate more about this? Do we need to create a contract that stores denom metadata?

contracts/hub/concentrated_vlp/src/testing/tests.rs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    #[test]
+    fn test_init() {
+        let mut deps = mock_dependencies_custom(&[]);
+        init(&mut deps);
+    }
+
+    #[test]
+    fn test_execute_swap() {
```
SlayerAnsh: Separate tests into modular structure.
1. Add Liquidity
2. Remove Liquidity
3. Swap

SlayerAnsh (changes requested):

contracts/hub/concentrated_vlp/src/testing/tests.rs
joemonem: At that time i thought so, but it doesn't seem necessary. That data can be stored in the vlp.

packages/euclid/src/msgs/concentrated_vlp.rs
joemonem: Router's already using `pool_config`, can't we just add a `concentrated` variant to that enum and handle it in the `match` statement?

SlayerAnsh: Yes

tests-integration/src/tests/factory.rs
```diff
@@ -680,6 +692,295 @@ fn run_add_liquidity(factory_chain_id: &str, router_chain_id: &str) {
...
+                },
+                slippage_tolerance_bps: 100,
+                timeout: None,
+                lp_token_name: "lpname".to_string(),
+                lp_token_symbol: "lpsymbol".to_string(),
+                lp_token_decimal: 6,
+                lp_token_marketing: None,
+                pool_config: PoolConfig::Concentrated {
+                    pair_type: PairType::Xyk {},
+                    asset_infos: vec![
```
SlayerAnsh: asset_infos is not needed because it doesn't matter if we use native or cw20. Token id will be the main identifier here.

tests-integration/src/tests/factory.rs
```diff
@@ -680,6 +692,295 @@ fn run_add_liquidity(factory_chain_id: &str, router_chain_id: &str) {
...
+                        AssetInfo::native(token_a.token.to_string()),
+                        AssetInfo::native(token_b.token.to_string()),
+                    ],
+                    token_code_id: 1,
```
SlayerAnsh: Do we need this?

tests-integration/src/tests/factory.rs
```diff
@@ -680,6 +692,295 @@ fn run_add_liquidity(factory_chain_id: &str, router_chain_id: &str) {
...
+                    factory_addr: factory_contract.address().unwrap().to_string(),
```
SlayerAnsh: We can remove this as we don;t have a factory implementation.

SlayerAnsh (changes requested):

joemonem:
The intention was a Concentrated VLP but the code turned out to be based on a Passive Concentrated Liquidity Pool.

PR #117, opened 2025-09-24
feat: Router release fee

joemonem:
# Pull Request
Release fee section in the [doc](https://docs.google.com/document/d/1J4PUS2l6SAhlcReFZveN65owyTy73YmvRD-t60V-Miw/edit?tab=t.0#heading=h.kppvj2ihxkto)

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor/Chore

## Related Issue

## Testing

Steps to test:

1. [Step 1]
2. [Step 2]
3. [Step 3]

## Checklist

- [ ] Self-review completed

[Add screenshots here]

## Additional Notes

### DIFF
--- a/contracts/hub/router/src/execute/base.rs
+++ b/contracts/hub/router/src/execute/base.rs
@@ -1,6 +1,6 @@
 use cosmwasm_std::{
-    ensure, from_json, to_json_binary, Addr, Binary, CosmosMsg, DepsMut, Env, IbcMsg, IbcTimeout,
-    MessageInfo, Response, SubMsg, Uint128, WasmMsg,
+    ensure, from_json, to_json_binary, Addr, Binary, CosmosMsg, Decimal, DepsMut, Env, IbcMsg,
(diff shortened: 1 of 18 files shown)


contracts/hub/router/src/execute/base.rs
```diff
@@ -425,6 +425,15 @@ pub fn execute_release_escrow(
             continue;
         }
 
+        let fee = RELEASE_FEES
+            .load(
+                deps.storage,
+                format!("{token}{}", sender.chain_uid.to_string()),
```
SlayerAnsh: Release fee will be based on released chain, not the sender chain. So if release is on injective then we take injective release fee, similarly if release fee is on base then we take base release fee for that token.

Also there will be a default release fee (for testing it can be 0.1%) if there is no release fee found for that token and chain

SlayerAnsh (changes requested):

PR #122, opened 2025-11-05
feat: meta transaction contract

joemonem:
# Pull Request

## Description
Works like the relayer contract but only sends messages to router contract and the call data is a router `ExecuteMsg`.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor/Chore

## Related Issue

## Testing

Steps to test:

1. [Step 1]
2. [Step 2]
3. [Step 3]

## Checklist

- [ ] Self-review completed

[Add screenshots here]

## Additional Notes

### DIFF
--- a/contracts/hub/meta_transaction/src/contract.rs
+++ b/contracts/hub/meta_transaction/src/contract.rs
@@ -7,10 +7,7 @@ use euclid::error::ContractError;
 use euclid::msgs::meta_transaction::{ExecuteMsg, InstantiateMsg, QueryMsg, State};
 
 use crate::{
-    execute::{
-        execute_execute_authorized_transaction, execute_execute_meta_transaction,
-        execute_update_admin, execute_update_state,
-    },
+    execute::{execute_execute_meta_transaction, execute_update_admin, execute_update_state},
     query::{get_state, nonce_relayed},
     state::{AUTHORIZED_ADDRESSES, STATE},
 };
@@ -27,8 +24,6 @@ pub fn instantiate(
     msg: InstantiateMsg,
 ) -> Result<Response, ContractError> {
     let state = State {
-        relayer_pubkey: msg.relayer_pubkey,
-        relayer_address: msg.relayer_address.clone(),
         router_contract: msg.router_contract.clone(),
         admin: info.sender,
     };
@@ -37,7 +32,6 @@ pub fn instantiate(
     AUTHORIZED_ADDRESSES.save(deps.storage, &msg.authorized_addresses)?;
     Ok(Response::new()
         .add_attribute("method", "instantiate")
-        .add_attribute("relayer_address", msg.relayer_address)
         .add_attribute("router_contract", msg.router_contract))
 }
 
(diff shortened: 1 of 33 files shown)


contracts/hub/meta_transaction/src/contract.rs
```diff
@@ -0,0 +1,69 @@
...
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn instantiate(
+    deps: DepsMut,
+    _env: Env,
+    info: MessageInfo,
+    msg: InstantiateMsg,
+) -> Result<Response, ContractError> {
+    let state = State {
+        relayer_pubkey: msg.relayer_pubkey,
```
SlayerAnsh: There won't be relayer pubkey or address in meta transaction contract as any user can call this

contracts/hub/meta_transaction/src/contract.rs
```diff
@@ -0,0 +1,69 @@
...
+    mut deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    msg: ExecuteMsg,
+) -> Result<Response, ContractError> {
+    match msg {
+        ExecuteMsg::ExecuteMetaTransaction(msg) => {
+            execute_execute_meta_transaction(&mut deps, &env, &info, msg)
+        }
+        ExecuteMsg::ExecuteAuthorizedTransaction(msg) => {
```
SlayerAnsh: This won't be needed

contracts/hub/meta_transaction/src/execute.rs
```diff
@@ -0,0 +1,184 @@
...
+        &Uint128::from(env.block.height),
+    )?;
+
+    // Ensure the timestamp is not exceeded
+    ensure!(
+        env.block.time <= Timestamp::from_seconds(meta_transaction.expiry),
+        ContractError::new("Timestamp limit exceeded")
+    );
+
+    let verified = verify_signature(
```
SlayerAnsh: Verify based on chain type. Message provided in the MsgSignData will have
1. signer address on source chain (0x123123... for evm and euclid12323.... or osmo123123... for cosmos)
2. chain uid of source chain which will be used to get chain type (cosmos or evm). Query router for chain type
3. Pubkey of signer to verify signature
4. Msg that we need to process like swap, withdraw, transfer.

Verify should check of address actually matches using the derive address method and also signature is valid or not

contracts/hub/meta_transaction/src/execute.rs
```diff
@@ -0,0 +1,184 @@
...
+            )
+        }
+    );
+    // Ensure the nonce is not used
+    ensure!(
+        !NONCES.has(deps.storage, meta_transaction.nonce.clone()),
+        ContractError::new(format!("Nonce already used: {}", meta_transaction.nonce).as_str())
+    );
+    // Save the nonce
+    NONCES.save(
```
SlayerAnsh: Noce will be saved against sender key (chainuid:address) as each user will have their own nonce

SlayerAnsh (changes requested):

SlayerAnsh (commented):
LGTM! Lets add some test for usage reference.

PR #123, opened 2025-11-21
ADR-001: Use Vouchers Exclusively

joemonem:
# Title
Migrating Exclusively to Vouchers

## Status
Proposed

## Context
Having both Native and Vouchers in the system creates confusion for users. It makes keeping track of balances more difficult, steepens the learning curve for using our system, hampers performance and overall makes the whole system more complex.

## Decision
Use Vouchers exclusively. Users will need to have a voucher balance in our system to use it, and our system will only operate with Vouchers.

## Consequences
Will require a significant refactoring of our system, which will delay audits and our launch date.
Our system would require more trust from users since they'd have to convert their assets to our Vouchers from the beginning.
Makes security an even bigger concern because all the user funds will be custodied by our system.

Our system's performance will greatly improve.
The system will become easier to use and less complex: Balances will become clear since they'll no longer be separated by vouchers and non-vouchers.

### DIFF
--- a/ADRs/001-migrating-exclusively-to-vouchers.md
+++ b/ADRs/001-migrating-exclusively-to-vouchers.md
@@ -1,0 +1,20 @@
+# Title
+Migrating Exclusively to Vouchers
+
+## Status
+Proposed
+
+## Context
+Having both Native and Vouchers in the system creates confusion for users. It makes keeping track of balances more difficult, steepens the learning curve for using our system, hampers performance and overall makes the whole system more complex.
+
+## Decision
+Use Vouchers exclusively. Users will need to have a voucher balance in our system to use it, and our system will only operate with Vouchers.
+
+## Consequences
+Will require a significant refactoring of our system, which will delay audits and our launch date.
+Our system would require more trust from users since they'd have to convert their assets to our Vouchers from the beginning.
+Makes security an even bigger concern because all the user funds will be custodied by our system.
+
+Our system's performance will greatly improve.
+The system will become easier to use and less complex: Balances will become clear since they'll no longer be separated by vouchers and non-vouchers.
+


rmehta-cpu:
Will exclusive use of vouchers only create a hurdle with aggregator and DEX integrations?
We need to keep in mind maximum volume of ours will come from B2B and not retail.

daniel-wehbe:
I think moving to voucher only (and we stop calling it voucher but more of a deposit to platform) is much nicer for UI/UX and users. There are several places where our platform feels a bit clunky due to this dual token model we are using where other platforms seem smoother on how they operate for users.

daniel-wehbe:
> Will exclusive use of vouchers only create a hurdle with aggregator and DEX integrations? We need to keep in mind maximum volume of ours will come from B2B and not retail.

Would it be possible to only do this for our platforms and not entirely restrict the use of native?

rmehta-cpu:
> > Will exclusive use of vouchers only create a hurdle with aggregator and DEX integrations? We need to keep in mind maximum volume of ours will come from B2B and not retail.
> 
> Would it be possible to only do this for our platforms and not entirely restrict the use of native?

I’m trying to understand the purpose of vouchers. Were they designed to be the primary way to transact, or mainly a fallback when we had low liquidity and needed users to trade with a platform native intermediate token?

Even on testnet, many users complained when they received vouchers instead of native tokens. On mainnet, with deeper liquidity and more expectations, I don't see strong reasons why users would prefer to hold vouchers. Backlash would likely be higher.

If we can deliver sub 3 second swaps using native tokens, I believe using native directly is the best approach with the least integration friction.

UI can be improved, but I expect that 90%+ of Euclid’s mainnet volume will come through integrations, not our own front end. Because of that, we should prioritize whatever minimizes friction for DEXs, aggregators, and partners.

gachouchani1999:
I believe both are equally important. There is multiple reasons to integrate Euclid, either for price speculation or for receiving an asset with a purpose to use or store it. 

When you look at Binance, you are basically minting vouchers and using them directly on the wallet. Doing only vouchers does not make sense if you do not have an easy wallet to handle all these vouchers. 

The main issue with our users is that in Metamask, they are unable to touch these vouchers, hence moving to fully vouchers before this is handled in 2026 does not makes sense. 

Having Voucher only apps such as Euclid Launch is a great push in the right direction as they can handle balances directly from a portfolio section and withdraw/deposit assets when needed. For integrations, they might require a deposit & swap in one go and hence only vouchers could impede this process. 

At the end of the day, our goal is for users to trade exclusively through our liquidity layer and manage their portfolio through integrated apps and have more integrations in wallets, this is where voucher only makes sense. 

At that time, vouchers are then renamed to a deposit in the "system" which is completely permissionless, which makes us a permissionless decentralized Binance, this is where our moat is.

daniel-wehbe:
> Even on testnet, many users complained when they received vouchers instead of native tokens. On mainnet, with deeper liquidity and more expectations, I don't see strong reasons why users would prefer to hold vouchers. Backlash would likely be higher.

The reason for the above is the use of both native and voucher and the concept of voucher which is not very clear for a user as why they would need it. My comments are mostly based on my extensive use of our applications and my frustration with the dual system of tokens which many of our users will face. If our testnet only had one token type (either native or voucher which would not be called voucher anymore), then the complaints would have been much much less as it would be less confusing. I like the idea of voucher only, as it is faster not only for swaps but also querying data from multiple chains (balance fetching and other processes which are not ideal currently with our setup as loading times are pretty long especialy with passport connection). My takes are not really based on integrations which is a huge part as well. So we wouldnt want to do anything to make that process harder.

PR #125, opened 2026-01-05
feat: Voucher & VLP Pause/Unpause

joemonem:
# Pull Request

## Description
Add pause/unpause functions in vlp and virtual balance contracts which can only be called by their admin.
Added queries for the paused tokens in virtual balance.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor/Chore

## Testing
Created unit tests

## Checklist

- [ ] Self-review completed

## Additional Notes
We can discuss if the router should be aware of the paused vlps/vouchers to fail certain messages instead of sending them and then failing in the destination contract.
After discussion, decided that all functionality should be handled within each contract and not called by the router contract


contracts/hub/virtual_balance/src/state.rs
```diff
@@ -18,3 +21,24 @@ pub struct Allowance {
 // Allowance is stored as a map of balance key to allowance. It allows another user to spend on behalf of the owner.
 // Only 1 allowance per balance key is allowed at a time.
 pub const ALLOWANCES: Map<SerializedBalanceKey, Allowance> = Map::new("allowances");
+
+// A map of ChainUid and TokenId to the blockheight when the token was paused
+pub const PAUSED_TOKENS: Map<(ChainUid, String), u64> = Map::new("paused_tokens");
```
SlayerAnsh, line 24: Should we do (String, ChainUid) as it will be easy to get all the tokens that are paused and then check for chains where its paused if we flip the tupple?

contracts/hub/virtual_balance/src/execute.rs
```diff
@@ -343,6 +357,9 @@ pub fn execute_approve(
         ContractError::Unauthorized {}
     );
 
+    // Check if token is paused
+    token_pause_check(deps.storage, vsl_chain_uid.clone(), msg.token_id.clone())?;
```
SlayerAnsh: Approvals should be fine, it doesn't transfer asset

contracts/hub/virtual_balance/src/state.rs
SlayerAnsh: Current structure also looks good as we can query all the tokens disabled for a single chain easily.

contracts/hub/virtual_balance/src/tests.rs
```diff
@@ -339,6 +337,80 @@ mod tests {
                 recipient_key.clone().to_serialized_balance_key(),
             )
             .unwrap_err();
+
+        // Pause token
+        let pause_msg = ExecuteMsg::PauseToken {
```
SlayerAnsh, line 371: Can we move this to a decicated pause test?

packages/euclid/src/pool/pool_functions.rs
```diff
@@ -217,6 +218,7 @@ pub fn update_fee(
     recipient: Option<CrossChainUser>,
 ) -> Result<Response, ContractError> {
     let mut state = state_storage.load(deps.storage)?;
+    ensure!(!state.paused, ContractError::ContractPaused {});
```
SlayerAnsh: We don't need pause here

SlayerAnsh (changes requested):
Can we add integration test for paused swap using
1. Paused VLP - Swap will error if vlp is included in a route
2. Paused Token - Error swap route that includes the token

PR #137, opened 2026-03-03
Feat: Added Flat Release Fee, Custom Chain Timeout, All Balance Query to vcoin and debug test for add liquidity slippage issue

SlayerAnsh:
# Pull Request

## Description

Updated release fee from bps to flat fee to allow easy and efficient release calculation.
Added custom chain timeout for releases so it can be customized to higher value for slow chains.
Added new queries to query all balances and token balance in virtual balance contract
Added test to ensure proper error message is generated for slippage exceed during add liquidity instead of pure codespace wasm error

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor/Chore

## Related Issue

## Testing

Steps to test:

1. [Step 1]
2. [Step 2]
3. [Step 3]

## Checklist

- [x] Self-review completed

[Add screenshots here]

## Additional Notes

### DIFF
--- a/contracts/hub/router/src/state.rs
+++ b/contracts/hub/router/src/state.rs
@@ -1,5 +1,5 @@
 use cosmwasm_schema::cw_serde;
-use cosmwasm_std::{Addr, Decimal, Uint128};
+use cosmwasm_std::{Addr, Uint128};
 use cw_storage_plus::{Item, Map};
 use euclid::{
     chain::{Chain, ChainUid},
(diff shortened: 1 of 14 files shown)


contracts/hub/router/src/state.rs
```diff
@@ -70,5 +70,8 @@ pub const PENDING_RELEASE_VOUCHER: Map<String, PendingReleaseVoucher> =
 pub const FUNDS_INFO: Item<(PairWithDenomAndAmount, u64)> = Item::new("funds_info");
 
 /// The key is TokenID_ChainUID
-pub const RELEASE_FEES: Map<(Token, ChainUid), Decimal> = Map::new("release_fees");
-pub const DEFAULT_RELEASE_FEE: Item<Decimal> = Item::new("default_release_fee");
+pub const RELEASE_FEES: Map<(Token, ChainUid), Uint128> = Map::new("release_fees");
+pub const DEFAULT_RELEASE_FEE: Item<Uint128> = Item::new("default_release_fee");
```
joemonem, line 74: This Item isn't being saved anywhere, only `.load()` is being called on it

joemonem: Also doesn't have query function that retrieves its value

contracts/hub/router/src/state.rs
joemonem, line 35: I think this should be renamed to fee recipients

SlayerAnsh: Initially it had some fee state so kept the same name

contracts/hub/router/src/state.rs
SlayerAnsh: Didn't added query because we can use raw state query for it. Assuming you are talking about DEFAULT_RELEASE_FEE only as RELEASE_FEES has a paginated query

joemonem: Could the default release fee cause problems since it's not tied to a specific token? 
For example introducing a 1 BTC release fee

SlayerAnsh: We will avoid it in most of the cases but it will cause issues with 18 decimal and 6 decimal tokens. We will handle that once we have better solution as its true for MIN_LIQUIDITY as well.

SlayerAnsh: We will have a service that continuously update default fee for all token based on their current price so this will be used rarely. Its present as a fallback if there is any delay between new token entry.

joemonem (approved):

PR #138, opened 2026-03-04
fix: use wrapping arithmetic for fee growth accumulators

crnbarr93:
# Pull Request

## Description

Replace all checked/saturating arithmetic with wrapping arithmetic for V3-style accumulators throughout the concentrated liquidity implementation. Fee growth and oracle cumulative values are unsigned integers designed to overflow (mod 2^256 or mod 2^128) — only differences between snapshots matter. Using checked or saturating arithmetic causes reverts or silent data corruption after the accumulators wrap.

**Changes:**

### position_math.rs
- `fee_growth_inside()`: all `checked_sub` → `wrapping_sub` for fee growth below/above/inside computation
- `fees_owed()`: removed `<=` early-return guard that masks wrapped deltas; `checked_sub` → `wrapping_sub` for delta computation
- New `accumulate_fee_growth()` helper: computes `global += fee * 2^128 / liquidity (mod 2^256)` with `wrapping_add`
- New `flip_fee_growth_outside()` helper: computes `global - outside (mod 2^256)` with `wrapping_sub` for tick crossing
- `TickInfo`: derive `Default`, use `unwrap_or_default()`

### contract.rs
- Swap loop fee accumulation: replaced inline `checked_add` with `accumulate_fee_growth()`
- Swap loop tick crossing: replaced inline `checked_sub` with `flip_fee_growth_outside()`

### oracle.rs
- New `advance_observation_cumulatives()` helper: uses `wrapping_add`/`wrapping_mul` for `tick_cumulative` (i128) and `wrapping_add` for `seconds_per_liquidity_cumulative_x128` (Uint256)
- `interpolate()`: `saturating_sub`/`saturating_add` → `wrapping_sub`/`wrapping_add` for both tick and spl deltas
- `write_observation()`: refactored to use `advance_observation_cumulatives()`

## Type of Change

- [x] Bug fix

## Testing

Steps to test:

1. `cargo test -p concentrated_vlp --lib`
2. Verify `fee_growth_inside_reverts_on_wrapped_global` passes (wrapping subtraction in fee_growth_inside)
3. Verify `fees_owed_guard_masks_wrapped_fee_growth` passes (wrapping delta in fees_owed)
4. Verify `accumulate_fee_growth_table` passes (8 cases including overflow wrapping and rounding)
5. Verify `flip_fee_growth_outside_table` passes (6 cases including wrapped overflow and double-flip identity)
6. Verify `advance_observation_cumulatives_table` passes (7 cases including i128 and Uint256 wrapping)
7. Verify `interpolate_wraps_correctly` passes (interpolation across wrapped boundaries)

## Checklist

- [x] Self-review completed

## Additional Notes

All wrapping arithmetic changes follow Uniswap V3's design where accumulators are monotonically increasing values that overflow by design. Only the *difference* between two snapshots is meaningful, and wrapping subtraction always produces the correct delta regardless of absolute values.


contracts/hub/concentrated_vlp/src/math/oracle.rs
```diff
@@ -97,25 +119,24 @@ fn interpolate(left: &Observation, right: &Observation, target: u64) -> Observat
...
         .seconds_per_liquidity_cumulative_x128
-        .checked_sub(left.seconds_per_liquidity_cumulative_x128)
-        .unwrap_or_default();
+        .wrapping_sub(left.seconds_per_liquidity_cumulative_x128);
     let spl_interp = left
         .seconds_per_liquidity_cumulative_x128
-        .checked_add(
+        .wrapping_add(
             spl_delta
                 .checked_mul(Uint256::from(elapsed as u128))
```
gachouchani1999, line 135: not related to your PR but can then lead to divison by 0 on overflow as well? Better to use `mult_div` or anything similar?

crnbarr93: A `wrapping_add` can equal 0, I had claude double check and seems like we don't have any concerns with these changes. If that's what you're asking.

joemonem (approved):

PR #141, opened 2026-03-08
Add swap price limit and fix oracle ring buffer wraparound

crnbarr93:
# Pull Request

## Description

Three concentrated liquidity pool fixes:

### Swap price limit parameter (`sqrt_price_limit_x96`)

The swap simulation loop previously ran until `amount_remaining` was fully consumed, with no per-leg price protection. In multi-hop swaps, `min_token_out` is only checked on the final leg — intermediate legs could execute at arbitrarily bad prices if liquidity is thin, creating a sandwich attack vector.

This adds an optional `sqrt_price_limit_x96: Option<Uint256>` parameter to `run_swap_simulation` that halts the swap loop when the price reaches the caller-specified bound (matching Uniswap V3's `sqrtPriceLimitX96`). Validation ensures the limit is in the correct direction relative to the current price and within min/max sqrt ratio bounds. When a price limit is set, partial fills are allowed instead of reverting. The tick crossing logic correctly skips crossing when the step ends at the price limit rather than at a tick boundary.

The parameter is internal only — not yet wired to message types or execute handlers.

### Oracle ring buffer wraparound fix

The `observe()` function previously loaded ALL observations from storage via `OBSERVATIONS.range()`, including stale entries from previous ring buffer cycles that hadn't been overwritten yet. After the buffer wrapped around, these stale observations were included in the sorted time series, producing incorrect TWAP interpolation values for any protocol or integration relying on the oracle.

The fix replaces the full-scan approach with targeted reads using ring buffer metadata from `SLOT0` (`observation_index` and `observation_cardinality`). Only valid entries within the current ring are read, traversed in chronological order from oldest to newest without requiring a sort.

### Misnamed swap_math test

Renamed `compute_swap_step_exact_output_vectors` to `compute_swap_step_exact_input_one_for_zero` — the test was calling the exact-input function, not an exact-output function, creating false coverage confidence.

## Type of Change

- [x] Bug fix
- [x] New feature

## Related Issue

CLP review findings

## Testing

Steps to test:

1. `cargo test -p concentrated_vlp` — all 27 tests pass
2. New oracle tests verify ring buffer behavior across 5 table-driven cases: linear buffer, wrapped buffer ignoring stale entries, single observation, wrapped buffer with interpolation, and full double-wrap
3. New swap limit tests verify: validation (8 table-driven cases for both directions), swap stopping at limit, full-consumption enforcement without limit, and partial fill with limit

## Checklist

- [x] Self-review completed

## Additional Notes

The `sqrt_price_limit_x96` parameter is not yet exposed in message types (`VlpSwapMsg`, `VlpSimulateSwapMsg`) or wired through the router/factory. This will be done in a follow-up once the parameter is ready to be surfaced to callers.


contracts/hub/concentrated_vlp/src/contract.rs
```diff
@@ -1498,3 +1550,314 @@ fn query_migration_status(deps: Deps) -> Result<MigrationStatusResponse, Contrac
         total_liquidity: state.total_lp_tokens,
     })
 }
+
+#[cfg(test)]
```
joemonem, line 1576: I think it would be better to move the tests to the `tests.rs` file

crnbarr93: Do we prefer a static test file over colocation? Wouldn't this file get egregiously large?

joemonem: Yeah good point

joemonem (approved):

PR #152, opened 2026-03-18
Instantiate position token contract from factory during init

joemonem:
## Overview
Factory contract now automatically instantiates the position token contract during initialization, eliminating the need for separate manual instantiation and registration steps.

## Changes
- Added `position_token_code_id` to factory state and instantiation message
- Factory now creates a submessage to instantiate position token during `instantiate()` with standardized name "Euclid Concentrated Positions" and symbol "EUPOS"
- Added `on_position_token_instantiate_reply()` handler to capture and store position token contract address
- Added `UpdatePositionTokenCodeId` variant to `ManageFactoryState` for updating code ID if needed
- Updated position token contract to expose mock helper for testing
- Simplified test setup by removing manual position token instantiation and registration steps
- Updated all factory instantiation calls and mocks to include `position_token_code_id` parameter


contracts/liquidity/factory/src/contract.rs
```diff
@@ -86,7 +91,27 @@ pub fn instantiate(
     STATE.save(deps.storage, &state)?;
     ADMIN.save(deps.storage, &EuclidAdmin::default(info.sender.clone()))?;
 
+    let init_position_token_msg = CosmosMsg::Wasm(WasmMsg::Instantiate {
+        admin: Some(info.sender.into_string()),
```
SlayerAnsh: We should use Admin state here for consistency. It will be same as info.sender so not a change in logic just a small nitpick

contracts/liquidity/factory/src/reply.rs
```diff
@@ -64,6 +65,34 @@ pub fn on_escrow_instantiate_reply(deps: DepsMut, msg: Reply) -> Result<Response
     }
 }
 
+pub fn on_position_token_instantiate_reply(
+    deps: DepsMut,
+    msg: Reply,
+) -> Result<Response, ContractError> {
+    match msg.result.clone() {
+        SubMsgResult::Err(err) => Err(ContractError::PoolInstantiateFailed { err }),
```
SlayerAnsh: Error message can be better

packages/euclid/src/msgs/factory/msg.rs
```diff
@@ -171,8 +172,8 @@ pub enum ManageFactoryState {
     UpdateRelayerAddress {
         relayer_address: String,
     },
-    UpdatePositionTokenContract {
-        position_token_contract: String,
+    UpdatePositionTokenCodeId {
```
SlayerAnsh, line 175: We need another method to RegisterPositionToken if we need to create new token in exisiting factory. This can be handled via migrate function as well but having a RegisterPositionToken method which registers if token is not present else error will make it easy to manage migration updates.

SlayerAnsh (changes requested):

packages/euclid/src/msgs/factory/msg.rs
joemonem: Addressed this by adding `ManageFactoryState::RegisterPositionToken`

contracts/liquidity/factory/src/reply.rs
```diff
@@ -64,9 +69,45 @@ pub fn on_escrow_instantiate_reply(deps: DepsMut, msg: Reply) -> Result<Response
...
+            let data = result.data.unwrap_or_default();
+
+            let instantiate_data: cw_utils::MsgInstantiateContractResponse =
+                parse_instantiate_response_data(&data).map_err(|res| ContractError::Generic {
+                    err: res.to_string(),
+                })?;
+
+            let position_token_address =
+                deps.api.addr_validate(&instantiate_data.contract_address)?;
+            POSITION_TOKEN_CONTRACT.save(deps.storage, &position_token_address)?;
```
SlayerAnsh, line 101: RegisterPositionToken can be called even if already instantiated during init (Low)                                                                                                                                           
                                                                                                                                                                                                                                  
  The guard POSITION_TOKEN_CONTRACT.may_load(deps.storage)?.is_none() prevents double-registration. But the reply handler on_position_token_instantiate_reply does a raw POSITION_TOKEN_CONTRACT.save() without checking if one   
  already exists. If by some edge case the reply fires twice (shouldn't happen, but defensive), it would silently overwrite. The guard in RegisterPositionToken covers the execute path, so this is low risk.

contracts/liquidity/factory/src/contract.rs
```diff
@@ -84,9 +89,30 @@ pub fn instantiate(
...
+        funds: vec![],
+        label: "position_token".to_string(),
+    });
 
     Ok(Response::new()
+        .add_submessage(SubMsg {
+            id: POSITION_TOKEN_INSTANTIATE_REPLY_ID,
+            msg: init_position_token_msg,
+            gas_limit: None,
+            reply_on: ReplyOn::Always,
```
SlayerAnsh: ReplyOn::Always vs ReplyOn::Success (Low)                                                                                                                                                                                    
                                         
  The submessage uses ReplyOn::Always, meaning the reply handler fires even on failure. The handler returns Err(ContractError::Reply { ... }) on failure, which will revert the whole transaction. This is correct behavior, but  
  ReplyOn::Success would be simpler since a failed submessage already reverts the parent by default. Not a bug, just unnecessary handling.

SlayerAnsh (changes requested):

contracts/liquidity/factory/src/tests.rs
```diff
@@ -71,463 +148,212 @@ mod tests {
...
-    //             token_1_liquidity: Uint128::new(1000),
-    //             token_2_liquidity: Uint128::new(2000),
-    //             liquidity_id: "liquidity_id_1".to_string(),
-    //             vlp_address: "vlp_address_1".to_string(),
-    //             pair_info: pair_info_1,
-    //         };
+    /// Identifies which address plays each role in a test case.
+    /// `Alice` = position metadata owner; `Bob` = a different address.
+    #[derive(Clone, Copy)]
+    enum Who {
```
SlayerAnsh, line 156: Adding a impl to this enum `to_address` can help simply some of the match statement below used to generate address. We can then directly use `nft_owner.to_address(deps);`

SlayerAnsh (changes requested):

SlayerAnsh (approved):

PR #155, opened 2026-03-22
feat: implement normalized voucher system and centralize escrow management

SlayerAnsh:
## Summary

Introduces a normalized voucher system where all token amounts are stored at 24-decimal precision using Uint256. Centralizes escrow management by moving ESCROW_BALANCES from the router contract to virtual_balance.

### Core Architecture
- All voucher balances stored at 24-decimal precision (Uint256)
- Escrow tracking moved from router to virtual_balance
- Token metadata registry for decimals and types to support normalization
- Breaking API changes: ExecuteMint, ExecuteBurn, VoucherReceive use Uint256 with source chain metadata

### Contract Migrations (new)
- **virtual_balance**: migrates BALANCES→VOUCHER_BALANCES with 24-dec normalization, seeds TOKEN_METADATA from MigrateMsg, queries router GetAllEscrows to seed ESCROW_BALANCES. Fails if any balance token lacks metadata.
- **cp_vlp / stable_vlp**: normalizes pool reserves by querying virtual_balance for token decimals and verifying normalized reserves match VLP voucher balance in virtual_balance. LP supply unchanged (proven safe by math tests).
- **router**: deprecated GetAllEscrows query added for migration use. No data migration needed (Uint128→Uint256 backward compatible).
- **escrow / factory**: no data migration needed (Uint128→Uint256 backward compatible, TokenType Option fields default to None).

### Migration Deployment Order
1. Deploy updated virtual_balance (with token_metadata in MigrateMsg)
2. Deploy updated cp_vlp / stable_vlp (queries virtual_balance for decimals)
3. Deploy updated router (no data migration, just code upgrade)
4. Deploy updated escrow, factory (no data migration)

### Shared Utilities
- `packages/euclid/src/utils/migration.rs`: query_token_decimals and query_voucher_balance helpers (used by VLP migrations)
- `packages/euclid/src/normalize.rs`: normalize_token_to_voucher and normalize_voucher_to_token with decimals > 24 rejection guard

### LP Token Safety
- Existing LP holders' proportional ownership preserved after reserve normalization
- Proven by 49 unit tests covering normalization, round-trips, migration scenarios, asymmetric decimal pairs, MINIMUM_LIQUIDITY boundaries, and Decimal256 precision analysis
- Value per LP scales by 10^(24-decimals), denormalized back = same raw amount

### Deprecated State Cleanup
- Router's ESCROW_BALANCES: definition kept (storage compat), all test fixtures and dead imports removed
- Router's TOKEN_DENOMS: deprecated, left in storage
- Test fixtures updated from `decimals: Some(24)` (no-op) to `Some(6)` to exercise real normalization

### Normalization Guard
- Dust amounts in `_release_voucher` that truncate to zero raw tokens after normalization are gracefully skipped (returns zero release) instead of erroring, preventing multi-recipient transaction aborts
- `normalize_token_to_voucher` and `normalize_voucher_to_token` reject decimals > 24 (base `normalize` fn still handles arbitrary conversions)
- Verified end-to-end across Native, IBC, and EVM chain types

### Ack Refund Path Verification
- Unit tests proving `ibc_ack_release_escrow` error path uses raw token units consistently (no unit mismatch, no double normalization)
- `PendingReleaseVoucher.total_amount` stores raw token units; `execute_mint` normalizes once

### Review Fixes (latest commit)
- Replace panic-on-overflow `.unwrap()` with error propagation in factory LP minting/burning (4 sites)
- Add Phase 2 idempotency guard (`may_load` skip) and strict single-metadata assertion in virtual_balance migration
- Remove dead unreachable `ensure!` in pool creation Ok(metadata) arm
- Remove dead `.try_into()` on already-Uint256 balance amount in swap
- Use voucher amount directly for voucher partner fees (skip unnecessary metadata query that could crash on unregistered chain)
- Switch VLP migrations from `.load()` to `.may_load()` for proper optional semantics
- Fix misleading `ZeroAssetAmount` error on `Limit::Dynamic` validation

### Normalize Unit Tests
- 74 unit tests in `packages/euclid/src/normalize.rs` covering: token-to-voucher, voucher-to-token, roundtrips, overflow, same-decimals identity, decimals > 24 rejection, Uint128 full roundtrip, math ops (exact addition, truncated division), voucher-exceeds-Uint128 scenarios
- All pool test cases converted to rstest `case::name` syntax for readable test output
- All test `.unwrap()` replaced with descriptive `.expect()` messages

## Type of Change

- [x] New feature
- [x] Breaking change
- [x] Refactor/Chore

## Testing

- `cargo test` (all tests pass, 96 router + 117 integration)
- `cargo test -p euclid -- normalize` (74 normalize tests)
- `cargo test -p euclid-pool` (110 pool tests with named cases)
- `cargo test -p euclid-pool -- voucher_lp_tests` (49 LP voucher tests)
- `cargo test -p virtual_balance -- migrate` (7 migration tests)
- `cargo test -p cp_vlp -- migrate` (6 migration tests)
- `cargo test -p stable_vlp -- migrate` (4 migration tests)
- `cargo test -p tests-integration --lib tests_reusable::voucher_release` (12 e2e normalization tests across Local/IBC/EVM)

### Test Audit and Hardening (latest commit)

Multi-agent review of test mathematical correctness and invariant coverage:

**Bugs fixed:**
- `MINIMUM_LIQUIDITY` test asserted 1000 instead of production value 1,000,000,000
- CW20 partner fee integration test compared wrong balance delta
- Fuzz runner snapshot type mismatch in `run_mixed_timed` (light vs full)

**Production guard added:**
- `MIN_AMP = 100` constant in `stable_math.rs` with validation in `compute_stable_swap`, `update_amp_factor`, and stable_vlp `instantiate`. Prevents Newton method underflow when `amp < 50` (leverage < 1)

**23 new pool unit tests:**
- CP k-invariant: `k_after >= k_before` verified across 8 parameter sets
- Fee conservation: `swap_amount + lp_fee + euclid_fee == amount_in` (8 cases)
- Independent golden values: 3 tests with hand-computed expected outputs (breaks circular assertions)
- Euclid fee exclusion: 4 cases verifying euclid_fee stays out of pool reserves

**Assertion tightening:**
- D preservation tolerance: 0.1% to 0.01% with `D_after >= D_before` directionality
- Multi-hop swap output: `> 0` to `>= min_amount_out`
- Normalization round-trip: weak inequality to exact truncation verification
- `state_sync` metadata fallback: silent passthrough to panic on missing metadata
- Post-swap integration assertions: LP unchanged + reserve direction checks

**Fuzz framework fixes:**
- `assert_transition` now includes detail in panic messages
- `RunStats::merge` added for `run_for_duration` stat aggregation


crnbarr93:
# PR #155 — Normalized Voucher System: In-Depth Review

**PR**: `feat/voucher-transmuter` → `development` | 168 files | +6609 / -2338
**Summary**: Stores all voucher balances at 24-decimal precision via `Uint256`, moves escrow tracking from router to virtual_balance, adds token metadata registry.

---

## CRITICAL / HIGH Severity

### 1. [HIGH] Stable swap uses `amount_in` (pre-fee) instead of `swap_amount` (post-fee)

**File**: `packages/pool/src/pool_functions.rs:575`

```rust
// Fee deducted:
let swap_amount = amount_in.checked_sub(lp_fee.checked_add(euclid_fee)?)?;

// Stable path: passes amount_in (BUG), not swap_amount
compute_stable_swap(
    &Decimal256::from_integer(amount_in / descale_factor),  // ← should be swap_amount
    ...
);

// CP path: correctly uses swap_amount
calculate_cp_swap(swap_amount, token_in_reserve, token_out_reserve)?;
```

The stable curve computes output on the **full input** (ignoring fee deduction), but `execute_swap` only adds `swap_amount + lp_fee` to reserves. Every stable swap leaks value from the pool. This is a pre-existing bug, but the PR carries it forward and the descaling amplifies the absolute magnitude.

### 2. [HIGH] Stable swap descaling truncates small amounts to zero, rescaling amplifies rounding errors by 10^6

**File**: `packages/pool/src/pool_functions.rs:573-583`

```rust
let descale_factor = Uint256::from(10u128.pow(6));
let swap_result = compute_stable_swap(
    &Decimal256::from_integer(amount_in / descale_factor),  // integer div truncates
    ...
);
(swap_result.return_amount * descale_factor, ...)  // amplifies any rounding by 10^6
```

If `amount_in < 10^6`, the descaled value is **zero**. Combined with Newton solver tolerance (`TOL = 1e-6`), a non-zero `return_amount` of 1 gets rescaled to `10^6` — tokens from nothing. Even for non-zero inputs, the integer div/mul round-trip loses up to `10^6 - 1` units per swap and amplifies solver rounding.

### 3. [HIGH] `min_amount_out` normalization uses arbitrary `metadata.first()` — wrong decimals possible

**File**: `contracts/hub/router/src/ibc/receive/swap.rs:100-108`

```rust
let asset_out_metadata_res = deps.querier.query_wasm_smart(
    virtual_balance_address.to_string(),
    &QueryMsg::GetTokenMetadata { token_id: msg.asset_out.to_string(), pagination: None },
)?;
let asset_out_metadata = asset_out_metadata_res.metadata.first()...;
let normalized_min_amount_out = normalize_token_to_voucher(
    msg.min_amount_out, asset_out_metadata.token_type.get_decimals()?
)?;
```

`GetTokenMetadata` returns **all** metadata entries for a token across all chains/denoms. `.first()` picks whichever is in storage order. If the same token has different decimal configs on different chains (e.g., USDC 6-dec on Cosmos, 18-dec on EVM), this can normalize `min_amount_out` with the wrong decimals, making slippage protection 10^12x too loose or too tight. This enables sandwich attacks.

### 4. [HIGH] No migration of existing balances — all user funds become invisible

**File**: `contracts/hub/virtual_balance/src/migrate.rs`

The migration only handles admin state. The old `BALANCES` map (storage key `"balances"`) is never migrated to `VOUCHER_BALANCES` (key `"voucher_balances"`). After upgrade, every user's voucher balance reads as zero. Similarly:
- `ALLOWANCES` → `VOUCHER_ALLOWANCES`: not migrated
- Router's `ESCROW_BALANCES` → VB's `ESCROW_BALANCES`: not migrated
- `TOKEN_METADATA`: not seeded

This means **all existing balances are lost** until a separate migration step is performed. In-flight IBC packets will also fail because `TOKEN_METADATA` lookups return empty.

### 5. [HIGH] Mint/burn rounding asymmetry causes permanent value loss and escrow drift

**File**: `packages/euclid/src/normalize.rs:31` + `contracts/hub/virtual_balance/src/execute.rs`

- **Mint** (raw → 24-dec): lossless multiplication
- **Burn** (24-dec → raw): truncating division

In `_release_voucher`, the flow is: `voucher_amount` → `normalize_voucher_to_token` (truncates) → cap at escrow → `normalize_token_to_voucher` (back to exact multiple). The burn amount `voucher_release_amount` is the clean multiple, which may be **less** than the original `voucher_amount`. The difference is burned from the user's balance via the bookkeeping invariant check, but more voucher units are destroyed than escrow units released.

Over time, the escrow accumulates phantom balance that can never be redeemed. For 0-decimal tokens, the granularity is `10^24` — a user with `1.5 * 10^24` voucher units can only withdraw in multiples of `10^24`.

---

## MEDIUM Severity

### 6. [MEDIUM] `MINIMUM_LIQUIDITY` not scaled for 24-decimal precision

**File**: `packages/pool/src/pool_functions.rs:28`

```rust
pub const MINIMUM_LIQUIDITY: u128 = 1000;
```

With 6-dec tokens, `1000` = `0.001` tokens — meaningful first-depositor attack protection. With 24-dec normalized amounts, `1000` = `10^-21` tokens — **zero effective protection**. The classic LP inflation attack (tiny deposit → donate → grief future depositors) is now trivially feasible.

**Fix**: Scale to `10^15` or `10^18`.

### 7. [MEDIUM] `remove_liquidity` uses `checked_mul_ceil` — rounds up in user's favor

**File**: `packages/pool/src/pool_functions.rs` (remove_liquidity)

```rust
let token_1_liquidity = total_reserve_1.checked_mul_ceil(lp_share)?;
```

Ceiling rounds mean each LP withdrawer gets slightly more than their share. The last withdrawer's tx will fail when reserves are insufficient. Standard AMM practice is `checked_mul_floor`.

### 8. [MEDIUM] Deregistered tokens can still be minted

**File**: `contracts/hub/virtual_balance/src/execute.rs` — `execute_mint`

`execute_deregister_token_metadata` sets `allowed = false`, but `execute_mint` never checks the `allowed` field. Users can still deposit tokens using deregistered denoms. The burn path checks `allowed` (via router's `_release_voucher`), but the mint path does not.

### 9. [MEDIUM] `query_token_balances` has multiple panicking `.unwrap()` calls

**File**: `contracts/hub/virtual_balance/src/query.rs:150-167`

```rust
.for_each(|res| {
    let ((chain_uid, _, _), balance) = res.unwrap();  // panics
    ...existing_balance.checked_add(balance).unwrap();  // panics
});
...
ChainUid::create(chain_uid.clone()).unwrap()  // panics
```

If called from within a contract execution via `query_wasm_smart`, a panic aborts the entire transaction. This violates the codebase convention: "No `.unwrap()` in smart contracts."

### 10. [MEDIUM] `Uint128::try_from(amount).unwrap()` in `create_transfer_msg` / `create_escrow_msg`

**File**: `packages/euclid/src/token.rs:482,502,553`

These functions convert `Uint256` → `Uint128` with `.unwrap()`. If a denormalized amount somehow exceeds `u128::MAX` (possible for tokens with 0 decimals at high supply), the contract panics. These are called in the escrow release path — a panic here means stuck funds.

### 11. [MEDIUM] Escrow uses raw amounts, vouchers use normalized — cross-domain invariant never enforced

Escrow balances are in raw token units, voucher balances in 24-dec normalized units. No invariant check verifies that `sum(escrow_raw * 10^(24-dec))` = `sum(voucher_balances)` for each token. The two ledgers can drift silently via rounding (see #5) or bugs in caller code.

### 12. [MEDIUM] 0-decimal tokens create extreme normalization factors without validation

**File**: `contracts/hub/virtual_balance/src/execute.rs` — `execute_register_token_metadata`

Tokens with `decimals = 0` are accepted. The normalization factor becomes `10^24`. A single raw unit normalizes to `10^24` voucher units. Combined with `Decimal256::from_integer()` in stable math (which internally stores `value * 10^18`), values overflow the `Uint256` backing of `Decimal256` for moderate amounts.

**Fix**: Enforce `decimals >= 6` (or some reasonable minimum).

---

## LOW Severity

### 13. [LOW] `_deduct_allowance` defaults spender to `from` (the owner)

When no allowance exists, the default `VoucherAllowance { spender: from.clone(), ... }` combined with `ensure!(allowance.spender == sender)` would pass if `sender == from`. The zero amount check catches it, but the semantics are fragile. Should return an explicit `NoAllowance` error.

### 14. [LOW] `to_uint128_with_precision` returns `Uint256` — misleading name

**File**: `packages/euclid/src/utils/math.rs`

The trait method name says `uint128` but now returns `Uint256`. Maintenance hazard.

### 15. [LOW] Forwarding package has duplicate `TokenType` without decimals

**File**: `packages/forwarding/src/msgs/common_old.rs`

A separate `TokenType` definition exists without the `decimals` field. This creates a type divergence that could cause deserialization issues if forwarding contracts receive messages with the new `TokenType` format.

### 16. [LOW] `execute_approve` does not check owner has sufficient balance

Standard ERC-20 behavior, but in a cross-chain context where failed txs have IBC timeout costs, consider adding an optional balance check.

---

## Test Coverage Gaps

1. **No tests for 0-decimal tokens** — the most extreme normalization case
2. **No tests for non-round voucher burns** — amounts that truncate during denormalization
3. **No migration test** for `BALANCES` → `VOUCHER_BALANCES` (because migration logic doesn't exist)
4. **No test for cross-chain decimal mismatch** — mint from 6-dec chain, release to 18-dec chain
5. **No overflow boundary tests** — values near `Uint256::MAX / 10^24` that could overflow during normalization
6. **No end-to-end test** for full deposit → normalize → swap → denormalize → release cycle
7. **`min_amount_out` normalization** — no test exercising `metadata.first()` with multi-chain tokens
8. **Stable swap descaling** — no test for amounts < `10^6` or near-zero inputs

---

## Summary

| Severity | Count | Key Theme |
|----------|-------|-----------|
| HIGH | 5 | Pool math bugs (stable fee, descaling), missing migration, rounding asymmetry, wrong metadata selection |
| MEDIUM | 7 | Unscaled constants, wrong rounding direction, missing validation, panicking unwraps |
| LOW | 4 | API confusion, naming, type divergence |

**Top 3 blockers before merge:**

1. **Fix stable swap fee bug** — pass `swap_amount` not `amount_in` to `compute_stable_swap`. This is a fund-loss bug.
2. **Implement balance migration** — without it, all existing user funds become invisible after upgrade.
3. **Fix `min_amount_out` metadata selection** — use chain-specific metadata instead of arbitrary `.first()`. Without this fix, slippage protection is broken for multi-chain tokens.

contracts/hub/virtual_balance/src/state.rs
```diff
@@ -1,22 +1,69 @@
...
     voucher::SerializedBalanceKey,
 };
 
 pub const STATE: Item<State> = Item::new("state");
 pub const ADMIN: Item<EuclidAdmin> = Item::new("admin");
 
-pub const BALANCES: Map<SerializedBalanceKey, Uint128> = Map::new("balances");
+pub const VOUCHER_DECIMAL: u32 = 24;
+#[deprecated(note = "BALANCES has been moved to VOUCHER_BALANCES")]
+pub const BALANCES: Map<SerializedBalanceKey, Uint256> = Map::new("balances");
```
joemonem: Shouldn't this be kept at `Uint128` since we're deprecating it?

contracts/hub/virtual_balance/src/state.rs
```diff
@@ -1,22 +1,69 @@
...
 
+// Voucher balances are stored as Uint256 to avoid precision loss.
+pub const VOUCHER_BALANCES: Map<SerializedBalanceKey, Uint256> = Map::new("voucher_balances");
+
+#[deprecated(note = "ALLOWANCES has been moved to VOUCHER_ALLOWANCES")]
 #[cw_serde]
 pub struct Allowance {
     pub spender: CrossChainUser,
-    pub amount: Uint128,
+    pub amount: Uint256,
```
joemonem: Same comment as above

joemonem:
## Migration Bug: `decimals_map` keyed by `token_id` only — ignores per-chain decimal variation

**Affected file:** `contracts/hub/virtual_balance/src/migrate.rs`, Phase 1 & Phase 3

### The problem

Phase 1 builds a `HashMap<String, u32>` keyed on `token_id` alone:

```rust
let mut decimals_map: HashMap<String, u32> = HashMap::new();
// ...
decimals_map
    .entry(metadata.token.to_string())
    .or_insert(metadata.token_type.get_decimals()?);
```

Phase 3 then looks up decimals by `token_id` only when normalizing old balances:

```rust
let token_id = &key.2;  // SerializedBalanceKey = (ChainUid, AnyChainAddress, TokenId)
let decimals = decimals_map.get(token_id)...
```

### Why this is wrong

`execute_register_token_metadata` in `execute.rs` enforces decimal consistency only *within* the same `(token_id, chain_uid, token_type)` storage key — not globally per `token_id`. The decimal equality guard fires only on re-registration of the exact same triple:

```rust
let token_metadata_storage = get_token_metadata_key(
    token.to_string(),
    chain_uid.clone(),
    token_metadata.token_type.clone(),
);
if token_metadata_storage.has(deps.storage) {
    ensure!(existing_decimals == decimals, ContractError::DecimalsMismatch { ... });
}
```

Nothing prevents `("usdc", "osmosis", Native{decimals: 6})` and `("usdc", "ethereum", EVM{decimals: 18})` from coexisting in `TOKEN_METADATA`. The system explicitly supports per-chain decimal variation by design.

With `or_insert`, whichever chain's metadata is encountered first in the `msg.token_metadata` vec wins for the entire migration. All USDC balances — regardless of their actual origin chain — get normalized using that one decimal value. For any token with cross-chain decimal discrepancy this silently produces wrong voucher amounts with no error surfaced.

### Deeper gap

`SerializedBalanceKey = (ChainUid, AnyChainAddress, TokenId)` where `key.0` is the **holder's chain**, not the token's source chain. Since a user could hold vouchers sourced from multiple chains under a single balance key, the old `BALANCES` map doesn't record source-chain provenance at all, so there's no unambiguous per-balance decimal to pick.

### What needs clarification / fixing

1. **If Euclid enforces uniform decimals per `token_id` globally in practice** (i.e. USDC is always 6 decimals regardless of chain): document this invariant explicitly in `execute_register_token_metadata` and add an `ensure!` that rejects registration of the same `token_id` with different decimals across any chain. That makes `or_insert` safe and the migration correct by construction.

2. **If cross-chain decimal variation is intentional**: the migration needs a different strategy — e.g. require that all tokens in `BALANCES` have a single unambiguous decimal (failing loudly if not), or restructure the old balance storage to record source-chain provenance before this migration runs.

As written, the migration will produce silently incorrect normalizations for any token where per-chain decimal discrepancy exists.

SlayerAnsh:
In previous design, euclid enforced same decimals for all the tokens thats why all the balances have same decimals during migration. After migration contract will be able to support multiple decimals per token with normalization. Thats why migration only takes first token. Once migrated new structure will be enforced.

joemonem (approved):

contracts/hub/virtual_balance/src/migrate.rs
```diff
@@ -1,226 +1,396 @@
...
+    let mut metadata_count = 0u64;
+    for metadata in &msg.token_metadata {
+        let key = get_token_metadata_key(
+            metadata.token.to_string(),
+            metadata.chain_uid.clone(),
+            metadata.token_type.clone(),
+        );
+        key.save(deps.storage, metadata)?;
+        decimals_map
+            .entry(metadata.token.to_string())
```
crnbarr93: **[MAJOR] decimals_map uses or_insert — first metadata entry wins for token_id**

In Phase 1 of the migration, `decimals_map.entry(token_id).or_insert(decimals)` means if the same token_id appears with different decimals on different chains (e.g., a token bridged with 18 decimals on one chain and 6 on another), only the first-encountered entry's decimals are used. Phase 3 then applies this single decimals value to ALL balances for that token_id regardless of which chain deposited them. If a token_id legitimately has different native decimals across chains, balances from the second chain would be normalized with the wrong scaling factor, permanently corrupting voucher balances.

The PR should either: (a) assert all decimals for the same token_id are equal, or (b) key the lookup by (token_id, chain_uid) and use the balance key's chain_uid during Phase 3.

contracts/hub/router/src/execute/token.rs
```diff
@@ -233,50 +259,56 @@ pub fn _release_voucher(
...
 
-    let escrow_key =
-        ESCROW_BALANCES.key((token.to_string(), recipient.recipient.chain_uid.clone()));
-    let escrow_balance = escrow_key
-        .may_load(deps.storage)?
-        .unwrap_or(Uint128::zero());
+    let normalized_token_amount =
+        normalize_voucher_to_token(voucher_amount, token_metadata.token_type.get_decimals()?)?;
+
+    if normalized_token_amount.is_zero() && !voucher_amount.is_zero() {
```
crnbarr93, line 279: **[MAJOR] Dust skip in _release_voucher silently discards burned vouchers**

When `normalized_token_amount.is_zero() && !voucher_amount.is_zero()`, the function returns `(vec![], Uint256::zero())` meaning no escrow release happens. However, the caller may have already burned the voucher_amount from the user's balance BEFORE calling _release_voucher (depending on the call path). If the burn happened first, the user loses their vouchers with no token release. If the burn happens after (based on the return value), then the zero return means no burn — but this needs verification across all call sites.

The PR description says 'Dust amounts... are gracefully skipped' but this could mask non-trivial losses for low-decimal tokens (e.g., a token with 0 decimals where 1 voucher unit < 10^24 would always truncate to 0).

packages/euclid/src/utils/migration.rs
```diff
@@ -0,0 +1,53 @@
...
+    token: &Token,
+) -> Result<u32, ContractError> {
+    let resp: GetTokenMetadataResponse = deps.querier.query_wasm_smart(
+        virtual_balance,
+        &VirtualBalanceQueryMsg::GetTokenMetadata {
+            token_id: token.to_string(),
+            pagination: None,
+        },
+    )?;
+    let first = resp.metadata.first().ok_or_else(|| {
```
crnbarr93, line 27: **[MINOR] query_token_decimals takes .first() — fragile assumption**

The migration utility `query_token_decimals` fetches `GetTokenMetadata` with `pagination: None` and takes `.first()`. While the VLP migrations call this only for tokens in a pair (which likely have consistent metadata), if multiple metadata entries exist for a token across chains with different decimals, `.first()` provides no deterministic ordering guarantee. The VLP migration uses this to decide the normalization factor for its reserves — if the wrong decimals are returned, reserves would be incorrectly scaled.

contracts/hub/virtual_balance/src/execute.rs
```diff
@@ -516,17 +591,20 @@ pub fn execute_normalize_balance_keys(
...
             // On collision, keep the higher allowance to avoid silently dropping approved value
-            if let Some(existing) = ALLOWANCES.may_load(deps.storage, normalized_key.clone())? {
-                let merged = Allowance {
+            if let Some(existing) =
+                VOUCHER_ALLOWANCES.may_load(deps.storage, normalized_key.clone())?
+            {
+                let merged = VoucherAllowance {
                     amount: existing.amount.max(allowance.amount),
                     spender: existing.spender,
+                    expires_at: None,
```
crnbarr93: **[MINOR] Allowance expiry lost during key normalization**

In `execute_normalize_balance_keys`, when merging two VoucherAllowances after case-normalizing keys, `expires_at` is hardcoded to `None`. This discards any existing expiry timestamp, effectively making merged allowances permanent. If a user had a time-limited allowance, normalizing their balance key removes the time constraint.

contracts/hub/router/src/ibc/ack_and_timeout.rs
```diff
@@ -164,11 +163,13 @@ pub fn ibc_ack_release_escrow(
                 cross_chain_user: refund_recipient.clone(),
                 token_id: token.to_string(),
             };
-            let mint_amount = amount.checked_add(pending_release_voucher.release_fee_amount)?;
+            let mint_amount = pending_release_voucher.total_amount;
```
crnbarr93, line 166: **[MINOR] ACK refund path: mint_amount changed from computed sum to stored total_amount**

The refund path in `ibc_ack_release_escrow` changed from computing `amount.checked_add(pending_release_voucher.release_fee_amount)` to using `pending_release_voucher.total_amount` directly. This relies on `total_amount` being set correctly when the PendingReleaseVoucher was created. If there's any code path that sets total_amount inconsistently (e.g., pre-fee vs post-fee), the refund would mint the wrong amount. The change is cleaner but removes a defensive cross-check.

contracts/hub/virtual_balance/src/execute.rs
```diff
@@ -31,19 +39,43 @@ pub fn execute_mint(
     // Reject mixed-case or empty addresses before mutating state
     msg.balance_key.cross_chain_user.validate()?;
 
+    let metadata_key = get_token_metadata_key(
```
crnbarr93, line 42: **[MINOR] execute_mint trusts caller-supplied token_source_chain_uid without verification**

In `execute_mint`, the caller provides `token_source_chain_uid` which is used as the key for both TOKEN_METADATA lookup and ESCROW_BALANCES increment. The router (which is the only authorized caller) is trusted, but if the router ever passes an incorrect chain_uid, escrow balances would be credited to the wrong chain. No cross-validation exists between the chain_uid in the balance_key and token_source_chain_uid.

tests-integration/src/tests_reusable/voucher_release.rs
```diff
@@ -0,0 +1,377 @@
+#![cfg(not(target_arch = "wasm32"))]
```
crnbarr93, line 1: **[MINOR] No integration test for multi-decimal token pairs (e.g., 6-dec + 18-dec pool)**

The voucher_release integration tests exercise normalization for individual tokens, but there's no visible test that creates a pool with asymmetric decimals (e.g., USDC/6 paired with WETH/18) and verifies that swap amounts, LP minting, and liquidity removal all produce correct results after normalization. The pool math now operates entirely in 24-dec space, so asymmetric pairs are the critical case to verify.

contracts/hub/virtual_balance/src/state.rs
```diff
@@ -1,16 +1,59 @@
...
+
 // Allowance is stored as a map of balance key to allowance. It allows another user to spend on behalf of the owner.
 // Only 1 allowance per balance key is allowed at a time.
+#[deprecated(note = "ALLOWANCES has been moved to VOUCHER_ALLOWANCES")]
 pub const ALLOWANCES: Map<SerializedBalanceKey, Allowance> = Map::new("allowances");
+
+pub const VOUCHER_ALLOWANCES: Map<SerializedBalanceKey, VoucherAllowance> =
+    Map::new("voucher_allowances");
+
+// Token Metadata is stored as a map of token id to token metadata.
```
crnbarr93: nit: incorrect comment — the mapping is a tuple of `(token, chain_uid, token_type_key)` to token metadata

contracts/hub/virtual_balance/src/state.rs
```diff
@@ -1,16 +1,59 @@
...
+
+pub fn get_token_metadata_key(
+    token: String,
+    chain_uid: ChainUid,
+    token_type: TokenType,
+) -> Path<TokenMetadata> {
+    TOKEN_METADATA.key((token, chain_uid, token_type.get_key()))
+}
+
+// Escrow balances are stored as Uint256 to avoid precision loss.
```
crnbarr93: nit: unnecessary comment

packages/euclid/src/msgs/virtual_balance/msg.rs
```diff
@@ -104,33 +137,115 @@ pub enum QueryMsg {
...
+    GetTokenMetadata {
+        token_id: String,
+        pagination: Option<Pagination<(ChainUid, TokenType)>>,
     },
+    #[returns(GetAllTokenMetadataResponse)]
+    GetAllTokenMetadata {
+        pagination: Option<Pagination<(String, ChainUid, String)>>,
+    },
+    #[returns(GetTokenRegisteredResponse)]
+    GetTokenRegistered { token_id: String },
```
crnbarr93: nit: naming — unsure what this query provides from the name alone

contracts/hub/virtual_balance/src/migrate.rs
```diff
@@ -1,226 +1,396 @@
...
+    let escrow_resp: AllEscrowsResponse = deps
+        .querier
+        .query_wasm_smart(state.router.to_string(), &RouterQueryMsg::GetAllEscrows {})?;
+
+    let mut escrow_count = 0u64;
+    for escrow in &escrow_resp.escrows {
+        let token_id = escrow.token.to_string();
+        let chain_uid = escrow.chain_uid.clone();
+
+        let metadata_entries: Vec<_> = TOKEN_METADATA
```
crnbarr93: What do you think about adding an in-memory cache for this process? These `.prefix`/`range` methods can be costly, especially in a loop.

packages/euclid/src/utils/migration.rs
```diff
@@ -0,0 +1,53 @@
...
+    msgs::virtual_balance::msg::{
+        GetBalanceResponse, GetTokenMetadataResponse, QueryMsg as VirtualBalanceQueryMsg,
+    },
+    token::Token,
+    voucher::BalanceKey,
+};
+
+pub fn query_token_decimals(
+    deps: &DepsMut,
+    virtual_balance: &str,
```
crnbarr93: nit: unsure what this param is — is it an address?

packages/euclid/src/utils/migration.rs
```diff
@@ -0,0 +1,53 @@
...
+            "No token metadata found for '{}' in virtual_balance. Migrate virtual_balance first.",
+            token
+        ))
+    })?;
+    first.token_type.get_decimals()
+}
+
+pub fn query_voucher_balance(
+    deps: &DepsMut,
+    virtual_balance: &str,
```
crnbarr93: nit: same as above — `holder_address` naming is unclear

tests-integration/src/tests_reusable/voucher_release.rs
```diff
@@ -0,0 +1,377 @@
...
+use euclid::msgs::router::query::QueryMsgFns as RouterQueryMsgFns;
+use euclid::msgs::virtual_balance::msg::QueryMsgFns as VirtualBalanceQueryMsgFns;
+use euclid::normalize::{normalize_token_to_voucher, normalize_voucher_to_token};
+use euclid::recipient::Recipient;
+use euclid::token::{Token, TokenType, TokenWithDenom};
+use euclid::voucher::BalanceKey;
+use factory::FactoryContract;
+use router::RouterContract;
+
+pub fn deposit_native_token(
```
crnbarr93, line 23: nit: function comments would be useful in this file

crnbarr93 (changes requested):
**Requesting changes.** The core normalization logic and escrow centralization are well-structured, but the migration has a correctness gap: `decimals_map.or_insert()` blindly uses the first-encountered decimals per token_id, meaning if any token is registered with different native decimals across chains, all voucher balances after that first entry would be normalized with the wrong scaling factor — a one-shot corruption with no rollback path. The dust-skip in `_release_voucher` is a secondary concern: it silently returns zero when voucher→token truncation produces nothing, but whether this burns user funds depends on whether the caller has already deducted the voucher balance upstream. If the protocol invariant guarantees uniform decimals per token_id across all chains, an explicit `ensure!` in Phase 1 asserting this would close the gap cheaply and make the migration safe to approve.

*Co-reviewed with Claude*

contracts/hub/virtual_balance/src/migrate.rs
SlayerAnsh: Fixed. Phase 1 now asserts decimals consistency per token_id — conflicting values fail the migration. Added unit test `test_migrate_fails_conflicting_decimals`.

contracts/hub/router/src/execute/token.rs
SlayerAnsh: Verified all call paths — burn happens INSIDE `_release_voucher` (line 349), after the dust check. If dust skip triggers, no burn occurs and callers either `continue` (transfer_voucher) or return zero (withdraw_voucher). No funds loss possible.

Could argue for returning an error instead of silent zero for better UX, but no correctness issue.

packages/euclid/src/utils/migration.rs
SlayerAnsh: Fixed. `query_token_decimals` now iterates all returned metadata entries and asserts they have the same decimals. Returns error on inconsistency.

contracts/hub/virtual_balance/src/execute.rs
SlayerAnsh: Resolved. `execute_normalize_balance_keys` removed entirely — migration was one-shot and already executed.

contracts/hub/router/src/ibc/ack_and_timeout.rs
SlayerAnsh: The old code computed `amount + release_fee_amount` which is algebraically equal to `total_amount` (set at line 321: `total_amount: release_amount` where release_amount is the pre-fee value). Equivalent, just cleaner. No code change needed.

contracts/hub/virtual_balance/src/execute.rs
SlayerAnsh: By design. Router is the sole authorized caller — standard CosmWasm trust model. Adding cross-validation would cost gas with no security benefit since the trust boundary is already at the router level.

tests-integration/src/tests_reusable/voucher_release.rs
SlayerAnsh: Added integration test `test_create_pool_token_registered_on_other_chain_fails` in factory_create_pool.rs — sets up two factory chains with asymmetric registration and verifies the router rejects via error ack.

contracts/hub/virtual_balance/src/state.rs
SlayerAnsh: Fixed.

contracts/hub/virtual_balance/src/state.rs
SlayerAnsh: Removed.

packages/euclid/src/msgs/virtual_balance/msg.rs
SlayerAnsh: Renamed to `GetTokenStatus` with response field `registered: bool`.

contracts/hub/virtual_balance/src/migrate.rs
SlayerAnsh: Done. Phase 2 now uses an in-memory `HashMap<(String, String), Vec<TokenMetadata>>` built from the Phase 1 `msg.token_metadata` input. No more storage range scans per escrow entry.

packages/euclid/src/utils/migration.rs
SlayerAnsh: Renamed to `virtual_balance_addr`.

packages/euclid/src/utils/migration.rs
SlayerAnsh: Same — renamed to `virtual_balance_addr`.

tests-integration/src/tests_reusable/voucher_release.rs
SlayerAnsh: Added doc comments to `deposit_native_token` and `withdraw_voucher`.

crnbarr93 (approved):

joemonem (commented):
### Remaining Partial Fix

**[MINOR] `migration.rs:query_token_decimals` — pagination default is 10, not unlimited**

The `GetTokenMetadata` query with `pagination: None` resolves to `Pagination::default()`, and the query handler does `.take(limit.unwrap_or(10))` — so at most 10 entries are returned. The decimal consistency check added in response to crnbarr93's comment iterates all _returned_ entries, but misses any beyond the 10th chain registration for a given token.

In practice, VLP tokens are registered on a small number of chains so this is unlikely to matter today. But the partial check gives false confidence. Fix: pass a large explicit limit (`Pagination::new(None, None, None, Some(u64::MAX))`) or iterate through all pages.

---

### Design Notes

**`execute_mint` event emission inconsistency**

`execute_burn` (`virtual_balance/execute.rs:152–170`) correctly emits `virtual_balance_change_event()` and `escrow_balance_change_event()` for off-chain indexers. `execute_mint` only emits raw `.add_attribute()` calls. Since both are inverse operations on the same state, indexers that subscribe to the structured event types will see burns but not mints consistently.

---

*Co-reviewed with Claude*

joemonem (approved):

PR #164, opened 2026-03-31
fix: enforce CrossChainUser address validation across contracts

SlayerAnsh:
# Pull Request

## Description

`CrossChainUser::validate()` enforces lowercase, non-empty addresses but was not being called in most execute handlers. This allowed mixed-case addresses to be stored as balance keys, creating duplicate entries for the same logical user.

This PR adds `validate()` calls across all execute handlers that accept CrossChainUser (virtual_balance, factory, orderbook_deposits, meta_transaction, claimer) and adds a paginated admin migration function (`NormalizeBalanceKeys`) to fix existing mixed-case balance keys in virtual_balance.

## Type of Change

- [x] Bug fix
- [x] New feature

## Related Issue

N/A

## Testing

Steps to test:

1. `cargo build` to verify all contracts compile
2. `cargo test` to verify all 11 new tests pass (3 in cross_chain_user, 8 in virtual_balance)
3. `cargo clippy` to verify no new warnings introduced

## Checklist

- [x] Self-review completed

## Additional Notes

The `NormalizeBalanceKeys` function uses cursor-based pagination (`start_after` + `limit`, default 100) to safely process large numbers of entries without hitting gas limits. It combines balances when both a mixed-case and lowercase entry exist for the same user. The admin should call it repeatedly until `normalized_count` returns 0.

### DIFF
--- a/contracts/hub/meta_transaction/src/execute.rs
+++ b/contracts/hub/meta_transaction/src/execute.rs
@@ -145,6 +145,7 @@ pub fn execute_meta_transaction(
         meta_transaction.data.signer_chain_uid.clone(),
         meta_transaction.data.signer_address.clone(),
     );
+    verified_sender.validate()?;
 
     for call_data in meta_transaction.data.call_data {
         let meta_receive = MetaReceive {
--- a/contracts/hub/meta_transaction/src/execute.rs
+++ b/contracts/hub/meta_transaction/src/execute.rs
@@ -145,6 +145,7 @@ pub fn execute_meta_transaction(
         meta_transaction.data.signer_chain_uid.clone(),
         meta_transaction.data.signer_address.clone(),
     );
+    // Reject mixed-case or empty addresses before dispatching calls
     verified_sender.validate()?;
 
     for call_data in meta_transaction.data.call_data {
--- a/contracts/hub/router/src/ibc/ack_and_timeout.rs
+++ b/contracts/hub/router/src/ibc/ack_and_timeout.rs
@@ -43,6 +43,9 @@ pub fn reusable_internal_ack_call(
         } => {
             let res = from_json(ack)?;
             let recipient = CrossChainUser::new(chain_uid, recipient.to_string());
+            // Reject mixed-case or empty addresses from IBC packet data
+            sender.validate()?;
+            recipient.validate()?;
             ibc_ack_release_escrow(deps, env, sender, amount, token, res, recipient, tx_id)?
         }
     };
--- a/contracts/hub/router/src/ibc/receive/base.rs
+++ b/contracts/hub/router/src/ibc/receive/base.rs
@@ -34,6 +34,8 @@ pub fn reusable_internal_call(
         ContractError::DeregisteredChain {}
     );
     let tx_id = msg.get_tx_id();
+    // Reject mixed-case or empty addresses from IBC packet data
+    msg.get_sender().validate()?;
 
     let mut response = match msg {
         RouterCrossChainExecuteMsg::RegisterDenom {
--- a/contracts/hub/virtual_balance/src/contract.rs
+++ b/contracts/hub/virtual_balance/src/contract.rs
@@ -5,8 +5,9 @@ use cw2::set_contract_version;
 use euclid::admin::EuclidAdmin;
 
 use crate::execute::{
-    execute_approve, execute_burn, execute_mint, execute_remove_zero_state_values,
-    execute_transfer, execute_update_admin, execute_update_router,
+    execute_approve, execute_burn, execute_mint, execute_normalize_balance_keys,
+    execute_remove_zero_state_values, execute_transfer, execute_update_admin,
+    execute_update_router,
 };
 use crate::query::{
     query_admin, query_all_balances, query_allowance, query_balance, query_state,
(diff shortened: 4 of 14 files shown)


joemonem:
## Code Review — PR #164: CrossChainUser Address Validation

### Overview

This PR fixes a real correctness bug: `CrossChainUser::validate()` (which enforces lowercase, non-empty addresses) was never called at entry points, allowing mixed-case addresses to be stored as separate balance keys for the same logical user. The fix adds `validate()` across six contract entry points and introduces an admin-gated `NormalizeBalanceKeys` migration to repair existing stale state.

---

### Correctness Issues

**[Medium] Allowance normalization silently drops value when a collision exists**

In `execute_normalize_balance_keys` (`execute.rs:489-496`):

\`\`\`rust
if !ALLOWANCES.has(deps.storage, normalized_key.clone()) {
    ALLOWANCES.save(deps.storage, normalized_key, &allowance)?;
}
// If a lowercase entry already exists, the mixed-case allowance is just dropped
normalized_count += 1;
\`\`\`

Balances are summed on collision (`checked_add`), but allowances are silently discarded if a lowercase key already exists. Allowances are not token balances so merging them isn't straightforward, but silently dropping a user's approved allowance is a loss of state. At minimum this case deserves a comment explaining the intentional behavior, or consider taking `max()` of the two values.

**[Medium] Missing `validate()` on `recipient` in `remove_liquidity_request`**

`factory/execute/pool.rs:337-346` accepts both `sender` and `recipient` as `CrossChainUser`. The PR adds `sender.validate()` to `execute_swap_request` and `execute_deposit_token`, but `remove_liquidity_request` validates neither field (it only calls `deps.api.addr_validate(&sender.address)`, which does not guarantee lowercase for cross-chain addresses, and doesn't touch `recipient` at all).

**[Low] Router IBC receive handlers not covered**

`router/src/ibc/receive/swap.rs`, `pool.rs`, and `token.rs` all accept `CrossChainUser` directly from IBC packet data without calling `validate()`. Since these are downstream of the factory (which now validates), the risk is lower, but it leaves a gap for any future direct router call path or if the router receives packets from a chain whose factory hasn't been upgraded yet.

---

### Design Concerns

**[Low] Shared `start_after` cursor for both BALANCES and ALLOWANCES**

`execute.rs:453`: The `start` binding is cloned and used for both map ranges in the same call. If a caller passes a `start_after` key to page through balances, that same cursor is applied to the allowance range, which has independent ordering. The intended workflow ("call until `normalized_count == 0`") works correctly with `start_after: None`, but the pagination parameter is misleading for callers who might try to use it for resumption. Consider splitting into two separate admin functions, or at least document this limitation clearly.

**[Low] `filter_map(|r| r.ok())` silently swallows deserialization errors**

`execute.rs:460,483`: Storage corruption errors are silently skipped. For a migration function running on production state, this may silently miss entries that failed to deserialize. Consider collecting errors and returning them, or at least emitting an attribute with a skipped-entry count.

**[Low] `normalized_count` conflates balance and allowance entries**

The return attribute `normalized_count` mixes normalized balance entries and normalized allowance entries. A caller who processes the storage in batches has no way to know if they're done with balances vs. allowances independently.

---

### What's Done Well

- **Placement of validation**: All `validate()` calls happen before any storage mutation — correct.
- **Auth gating**: `NormalizeBalanceKeys` correctly checks `general_admin`. Consistent with the existing `RemoveZeroStateValues` pattern.
- **Balance collision handling**: `BALANCES.update` with `checked_add` is the right approach — no balance is lost on key collision.
- **Test coverage**: 8 new tests in virtual_balance + 3 in the package — covers the happy path, collision merging, pagination, admin gating, and no-op case. The pagination test correctly uses `start_after` with the original (now-removed) key, which works because it's an exclusive bound.
- **Collect-before-mutate pattern**: Correctly avoids iterating and mutating storage simultaneously.
- **Minimal diff**: Additions are surgical; no unrelated refactoring.

---

### Summary

| Severity | Count |
|----------|-------|
| Medium   | 2     |
| Low      | 3     |

The core fix is correct and important. The two medium issues should be addressed before merging: the allowance value-drop on collision needs explicit handling or documentation, and `remove_liquidity_request`'s `recipient` field needs a `validate()` call. The router IBC handler gap is lower priority but worth a follow-up issue.

SlayerAnsh:
All review items addressed in commit 17208954:

**[Medium] Allowance normalization silently drops value on collision**
Changed to use `max()` of the two allowance amounts instead of silently dropping. The existing lowercase entry's spender is preserved. (`virtual_balance/src/execute.rs`, allowance normalization loop)

**[Medium] Missing validate() on recipient in remove_liquidity_request**
Added `sender.validate()` and `recipient.validate()` at the top of `remove_liquidity_request`. (`factory/src/execute/pool.rs:348`)

**[Low] Router IBC receive handlers not covered**
Added `get_sender()` method to `RouterCrossChainExecuteMsg` (`euclid_ibc/src/router_ibc.rs`) and a single `msg.get_sender().validate()?` call in the parent dispatch function `reusable_internal_call` (`router/src/ibc/receive/base.rs:38`), covering all 8 IBC handler variants in one place. Also added validation for `sender` and `recipient` in `ibc_ack_release_escrow` (`router/src/ibc/ack_and_timeout.rs:47`) which constructs a CrossChainUser from IBC ack packet data.

**[Low] Shared start_after cursor for BALANCES and ALLOWANCES**
Added doc comment on the function explaining the cursor limitation and recommending `start_after: None` for simplest usage.

**[Low] filter_map silently swallows deserialization errors**
Replaced `filter_map(|r| r.ok())` with explicit match that increments a `skipped_errors` counter. The count is returned as a response attribute.

**[Low] normalized_count conflates balance and allowance entries**
Split into separate `normalized_balances` and `normalized_allowances` attributes.

**Additional coverage found during review:**
- Added `from.validate()` for the optional `from: Option<CrossChainUser>` param in `factory/src/execute/token.rs:execute_transfer_voucher`
- Added test `test_normalize_allowance_keeps_max_on_collision` covering the allowance max() collision behavior

joemonem (approved):

PR #177, opened 2026-04-08
refactor: rename lp_allocation to liquidity_delta for CLP

SlayerAnsh:
# Pull Request

## Description

Renames `lp_allocation` to `liquidity_delta` across the concentrated liquidity pool (CLP) remove liquidity path. The CP/Stable pools use fungible LP tokens, so `lp_allocation` is semantically correct there. However, concentrated liquidity positions use raw liquidity units (L) within a tick range, making `liquidity_delta` the accurate name. Also includes VlpResponse enhancement with optional `pool_key` for concentrated pools.

## Type of Change

- [x] Refactor/Chore

## Related Issue

N/A

## Testing

Steps to test:

1. `cargo clippy --all-targets --all-features`
2. `cargo test -p factory`
3. `cargo test -p router`

## Checklist

- [x] Self-review completed

## Additional Notes

The rename spans the full CLP remove liquidity flow: factory ExecuteMsg, factory state (`ConcentratedRemoveLiquidityRequest`), factory execute logic, IBC cross-chain message (`RouterCrossChainConcentratedRemoveLiquidityExecuteMsg`), router IBC receive handler, and integration test helpers. Non-concentrated (CP/Stable) paths retain `lp_allocation` unchanged.

### DIFF
--- a/contracts/liquidity/factory/src/ibc/ack_and_timeout.rs
+++ b/contracts/liquidity/factory/src/ibc/ack_and_timeout.rs
@@ -346,7 +346,7 @@ fn ack_concentrated_pool_creation(
                 data.position_id.u128(),
                 &crate::state::ConcentratedPositionMetadata {
                     owner: sender.clone(),
-                    pool_key: data.pool_key.clone(),
+                    pool_key: existing_req.pool_key.clone(),
                     liquidity: data.liquidity_delta,
                     vlp_address: data.vlp_address.clone(),
                 },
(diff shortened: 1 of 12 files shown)


contracts/liquidity/factory/src/ibc/ack_and_timeout.rs
```diff
@@ -52,7 +52,7 @@ pub fn reusable_internal_ack_call(
     match msg {
         RouterCrossChainExecuteMsg::RequestPoolCreation { tx_id, sender, .. } => {
             // Process acknowledgment for pool creation
-            let res: AcknowledgementMsg<PoolCreationResponse> = from_json(ack)?;
+            let res: AcknowledgementMsg<AddLiquidityResponse> = from_json(ack)?;
```
joemonem, line 55: What is the reason for this change?

packages/euclid_ibc/src/router_ibc.rs
```diff
@@ -225,7 +225,7 @@ pub struct RouterCrossChainConcentratedRemoveLiquidityExecuteMsg {
     pub sender: CrossChainUser,
     pub pool_key: PoolKey,
     pub position_id: Uint128,
-    pub lp_allocation: Uint128,
+    pub liquidity_delta: Uint128,
```
joemonem, line 228: Should we use `#[serde(alias = "lp_allocation")]` here?

contracts/liquidity/factory/src/ibc/ack_and_timeout.rs
SlayerAnsh: Any pool created has second step of adding liquidity to it. We added a check based on funds info for liquidity creation which added extra complexity, so now I have simplified it to always return AddLiqudityResponse which has all the data needed for ack.
So flow is always same for any type of pool
Creare Request -> VLP Created -> Liquidity Added -> Ack return Liqiidity response -> Factory mint liqiuidity tokens

packages/euclid_ibc/src/router_ibc.rs
SlayerAnsh: Explicitly renamed it to make sure we communicate it properly to developers that its different than lp_allocation

joemonem (approved):

PR #178, opened 2026-04-09
Centralize CLP position IDs and position token management

SlayerAnsh:
# Pull Request

## Description

Centralizes concentrated liquidity position ID generation in the Router (replacing the probabilistic FNV hash approach in the VLP), simplifies position state to use `chain_uid` instead of full `owner`/`pool_key`, and implements a common position token NFT contract managed by the Factory for minting, updating, and burning position tokens based on the VLP's `position_burned` flag. Adds comprehensive integration tests covering position lifecycle, authorization, pending/race conditions, and position ID behavior.

## Type of Change

- [x] New feature
- [x] Refactor/Chore

## Related Issue

N/A

## Testing

Steps to test:

1. `cargo test -p concentrated_vlp` (55 tests)
2. `cargo test -p factory` (21 tests)
3. `cargo test -p position_token` (46 tests)
4. `cargo test -p router` (115 tests)
5. `cargo test -p tests-integration -- clp::` (43 new integration tests)
6. `cargo test` for full suite including all integration tests

## Checklist

- [x] Self-review completed

### Router
- New `CLP_POSITION_ID_NONCE` counter and `CLP_POSITION_ID_VLP_MAP` for centralized position ID generation and VLP mapping validation
- Reply handler propagates `position_burned` flag from VLP response

### VLP (concentrated_vlp)
- `ConcentratedPosition` now stores `chain_uid: ChainUid` instead of `owner: CrossChainUser` + `pool_key: PoolKey`
- Removed `next_position_id()` and nonce functions
- Added `position_burned` flag to remove liquidity response (true only when both liquidity and owed fees are zero)
- Added `LegacyConcentratedPosition` struct for backward compatible migration

### Factory
- Single common position token contract (replacing per VLP approach)
- Mints NFT on first add liquidity, updates on subsequent adds/removes, burns only when `position_burned` is true
- Validates `vlp_address` from position info matches the pool's VLP on add/remove/collect
- `AddConcentratedLiquidityResponse` returned from helper with `position_id`, `liquidity_delta`, `used_token_1`, `used_token_2`

### Position Token
- Enhanced with `vlp_address` in `PositionInfo`, `UpdatePosition` execute message
- Comprehensive rstest based unit test suite (46 tests)

### Integration Tests (new, `tests-integration/src/tests_reusable/clp/`)
- **position_lifecycle.rs** (5 tests, 14 cases): Multi position creation, cross fee tier positions, partial remove metadata, full remove with fees, add/remove cycles with exact liquidity delta assertions
- **authorization.rs** (5 tests, 9 cases): Cross user rejection for add/remove/collect, cross pool position reuse rejection, nonexistent position ID rejection
- **pending_and_race.rs** (5 tests, 12 cases): Concurrent adds before ack delivery, remove then add sequential, double remove prevention, error ack liquidity restoration, fee collection during pending remove
- **position_id.rs** (4 tests, 11 cases): Sequential ID generation, position to VLP mapping validation, burned ID not reused, consistent ID across add operations


contracts/hub/concentrated_vlp/src/contract.rs
```diff
@@ -762,6 +748,7 @@ fn execute_remove_concentrated_liquidity(
         sender: remove_liquidity_msg.sender.clone(),
         vlp_address: env.contract.address.to_string(),
         pool_key: remove_liquidity_msg.pool_key,
+        position_burned,
```
joemonem: Is this variable necessary? If liquidity_after is zero, then this is always true.

contracts/hub/concentrated_vlp/src/contract.rs
```diff
@@ -697,10 +683,10 @@ fn execute_remove_concentrated_liquidity(
         (Uint128::zero(), Uint128::zero())
     };
 
-    if position.liquidity.is_zero()
+    let position_burned = position.liquidity.is_zero()
         && position.tokens_owed_0.is_zero()
```
joemonem: tokens owed are always zero if liquidity is zero, so we could remove those two checks and just leave `position.liquidity.is_zero()`

contracts/liquidity/position_token/src/state.rs
```diff
@@ -1,28 +1,14 @@
...
-    pub token_uri: Option<String>,
-}
+use euclid::msgs::position_token::{PositionInfo, State, TokenInfo};
 
 pub const STATE: Item<State> = Item::new("state");
+/// Map for token info
 pub const TOKENS: Map<&str, TokenInfo> = Map::new("tokens");
 
+/// Map for position info
+pub const POSITION_INFO: Map<&str, PositionInfo> = Map::new("positions");
```
joemonem: The map's name should be `position_info`

joemonem:
## Code Review: PR #178 — Centralize CLP Position IDs and Position Token Management

### Overview

This PR centralizes concentrated liquidity position ID generation from the VLP (probabilistic FNV hash) to the Router (sequential nonce), simplifies `ConcentratedPosition` state to `chain_uid` only, and introduces a single shared position token NFT contract managed by the Factory. The scope is substantial (~3800 additions), and the core architecture is sound, but there are several correctness and design issues worth addressing.

---

### Issues

#### Critical

**1. `get_clp_position_id` collision-scan loop is unnecessary and misleading — `router/src/state.rs`**

```rust
for _ in 0..iters {
    position_id = position_id.wrapping_add(1);
    if !CLP_POSITION_ID_VLP_MAP.has(deps.storage, position_id) {
```

Position IDs are sequential and are only removed from `CLP_POSITION_ID_VLP_MAP` when a position is burned. A simple `nonce + 1` will always be free unless ≥1000 consecutive positions were burned and IDs reassigned — but IDs are never reassigned. The loop exists to "avoid collisions," but there's no realistic collision scenario. This approach wastes gas on 1000 `has()` calls in a hypothetical failure case and adds a reachable `Err` path (`"No position id available after 1000 attempts"`) that will never trigger under correct operation, but will cause a very confusing failure if the nonce somehow wraps around `u128::MAX`. Replace with a plain increment — the current design makes the code look more defensive than it is.

**2. Swallowing query errors in `ack_add_concentrated_liquidity` — `factory/src/ibc/ack_and_timeout.rs`**

```rust
let existing_position: Result<PositionInfoResponse, _> = deps.querier.query_wasm_smart(...);
match existing_position {
    Ok(_position) => { /* update */ }
    Err(_err) => { /* mint */ }  // <-- catches ANY error
}
```

This treats *any* query failure (deserialization error, contract not found, out-of-gas) as "position doesn't exist yet" and proceeds to mint. If the position token contract has a bug, an old schema, or is temporarily unavailable, the factory will attempt to mint a token whose ID already exists, causing the entire ack tx to fail with a confusing `TokenAlreadyExist` error. The intent is clearly to distinguish "not found" from other errors, but a catch-all `Err(_err)` doesn't do that. Either use a specific `NotFound` error variant or check the error message — or better, include the position ID in the pending state and use it to determine mint-vs-update without a live query.

**3. Pool creation guard for position token removed — `factory/src/execute/pool.rs`**

```rust
// Removed:
POSITION_TOKEN_CONTRACT.load(...).or(Err(...))?;
```

Previously, requesting a concentrated pool creation would immediately fail if the position token contract wasn't registered. Now the failure surfaces in the IBC ack handler. This wastes gas and IBC fees on a request that was always going to fail, and produces a harder-to-diagnose error. The guard should be restored.

---

#### Medium

**4. `position_burned` redundancy — `concentrated_vlp/src/contract.rs`**

In `execute_remove_concentrated_liquidity`, the auto-collect block at lines 676–684 zeroes `tokens_owed_0` and `tokens_owed_1` whenever `liquidity.is_zero()`. So the three-condition check on lines 686–688 collapses to `position.liquidity.is_zero()`. The extra conditions are misleading — a reader might infer there's a scenario where liquidity is zero but fees are non-zero, which the code above prevents.

**5. Optimistic remove-then-refund has no confirmed timeout path for position token — `factory/src/execute/pool.rs` + `ack_and_timeout.rs`**

In `remove_concentrated_liquidity_request`, the factory immediately decrements the position token liquidity before the IBC round-trip. The error ack branch correctly calls `UpdatePosition` to refund. But check whether IBC **timeouts** go through the same `ack_remove_concentrated_liquidity` path with an `Error` variant — if timeouts are handled separately (e.g., via `execute_receive_packet_internal_callback` with a `None` ack), the refund message must also appear there. If not, a timed-out remove will leave the position token with stale (too-low) liquidity permanently.

**6. Misleading variable name — `factory/src/execute/pool.rs`**

```rust
let int128_liquidity_delta = Int256::from(liquidity_delta);
// ...
liquidity_change: -int128_liquidity_delta,
```

The variable is named `int128_liquidity_delta` but its type is `Int256`.

---

#### Low / Nits

**7. `Pair::display_name()` and `Pair::symbol()` are unused — `packages/euclid/src/token.rs`**
Neither method is called anywhere in the diff. If they're for future use, they can wait.

**8. `PoolConfig::to_string()` should use `..` instead of `_` field bindings — `packages/euclid/src/msgs/vlp/base.rs`**
```rust
PoolConfig::Stable { amp_factor: _ } => "stable".to_string(),
```
Prefer `PoolConfig::Stable { .. }` — idiomatic Rust for ignoring all fields.

**9. Test case naming in `position_token/src/execute.rs`**
The `#[case::int256_max]` and `#[case::int256_min]` labels are misleading — those cases actually test `u128::MAX` boundaries, not `Int256::MAX/MIN`.

**10. `factory/Cargo.toml` missing newline at EOF**

contracts/hub/concentrated_vlp/src/contract.rs
```diff
@@ -128,10 +127,9 @@ pub fn instantiate(
             },
         },
     )?;
-    initialize_position_nonce(deps.storage, env.contract.address.as_str())?;
     let initial_tick = msg.initial_tick.unwrap_or(0);
     ensure!(
-        initial_tick >= MIN_TICK && initial_tick <= MAX_TICK,
+        (MIN_TICK..=MAX_TICK).contains(&initial_tick),
```
crnbarr93: Is this not a pretty large operation? What was wrong with the comparison check?

contracts/hub/concentrated_vlp/src/contract.rs
```diff
@@ -497,23 +483,19 @@ fn execute_add_concentrated_liquidity(
...
             liquidity: Uint128::zero(),
-            pool_key: pool_key.clone(),
             fee_growth_inside_0_last_x128: Uint256::zero(),
             fee_growth_inside_1_last_x128: Uint256::zero(),
             tokens_owed_0: Uint128::zero(),
             tokens_owed_1: Uint128::zero(),
         });
 
-    if position.owner != sender {
+    if position.chain_uid != sender.chain_uid {
```
crnbarr93: nit: use `assert` to keep things consistent

contracts/hub/concentrated_vlp/src/contract.rs
```diff
@@ -645,12 +634,9 @@ fn execute_remove_concentrated_liquidity(
         .may_load(deps.storage, remove_liquidity_msg.position_id.u128())?
         .ok_or(ContractError::new("Position not found"))?;
 
-    if position.owner != remove_liquidity_msg.sender {
+    if position.chain_uid != remove_liquidity_msg.sender.chain_uid {
```
crnbarr93: nit: use `assert` to keep things consistent

contracts/hub/concentrated_vlp/src/contract.rs
crnbarr93: I think leaving it may be safer, in case there is some edge case we're unaware of, this could end up locking funds in the contract

contracts/hub/concentrated_vlp/src/contract.rs
SlayerAnsh: Yes, reverted the change

contracts/hub/concentrated_vlp/src/migrate.rs
```diff
@@ -91,9 +110,40 @@ pub fn migrate(deps: DepsMut, env: Env, msg: MigrateMsg) -> Result<Response, Con
     slot0.observation_cardinality = 1;
     slot0.observation_cardinality_next = 1;
 
-    let positions: Vec<(u128, ConcentratedPosition)> = POSITIONS
+    // Try loading positions in the current format first. If that fails (e.g. because
+    // the stored data still has the legacy `owner`/`pool_key` fields), fall back to
+    // deserializing as LegacyConcentratedPosition and converting.
+    let positions: Vec<(u128, ConcentratedPosition)> = match POSITIONS
```
crnbarr93: Do we have a unit test for this migration case?

SlayerAnsh: Removing migration logic as there won't be any old state we need migration

contracts/hub/concentrated_vlp/src/contract.rs
SlayerAnsh: Removed the separate variable and inlined `liquidity_after.is_zero()` directly in both the storage cleanup and the response struct. 35fb307e

contracts/hub/concentrated_vlp/src/contract.rs
SlayerAnsh: Simplified the check to just `liquidity_after.is_zero()` since the auto-collection block above already zeroes tokens_owed. 35fb307e

contracts/liquidity/position_token/src/state.rs
SlayerAnsh: Fixed, renamed storage key to `position_info`. 35fb307e

contracts/hub/concentrated_vlp/src/contract.rs
SlayerAnsh: Converted to `ensure!` for consistency with surrounding checks. 28a73b08

contracts/hub/concentrated_vlp/src/contract.rs
SlayerAnsh: Fixed in 28a73b08

SlayerAnsh:
### Resolution Summary

**Critical:**

1. **`get_clp_position_id` collision loop** — Keeping as-is. The loop is necessary for the wrapping case: when the nonce wraps around `u128::MAX`, it needs to skip occupied IDs and find freed slots from burned positions. We also now remove entries from `CLP_POSITION_ID_VLP_MAP` on burn (35fb307e), so the loop serves a real purpose.

2. **Swallowing query errors in `ack_add_concentrated_liquidity`** — Acknowledged. In practice the only realistic error is "not found" since the position token contract is validated earlier. Accepted risk for now.

3. **Pool creation guard restored** — Added `POSITION_TOKEN_CONTRACT` load check at the start of `execute_request_concentrated_pool_creation` so it fails fast before the IBC round-trip. 28a73b08

**Medium:**

4. **`position_burned` redundancy** — Simplified to `liquidity_after.is_zero()` and inlined (no separate variable). 35fb307e

5. **IBC timeout path for position token** — Already handled. The error ack branch in `ack_remove_concentrated_liquidity` restores position token liquidity via `UpdatePosition`. IBC timeouts flow through the same error ack path.

6. **Misleading variable name** — Renamed `int128_liquidity_delta` to `liquidity_delta_signed`. 28a73b08

**Low / Nits:**

7. **Unused `Pair::display_name()` and `Pair::symbol()`** — Removed. 28a73b08

8. **`PoolConfig::to_string()` field bindings** — Changed to `..` syntax. 28a73b08

9. **Test case naming** — Checked, all test cases have descriptive names. No `#[case::int256_max]` or `#[case::in]` found in current code.

10. **`factory/Cargo.toml` missing newline** — Fixed. 28a73b08

crnbarr93 (approved):

joemonem (approved):

PR #184, opened 2026-04-13
Fix: Escrow failure when creating new pool with 1 new token in CLP

SlayerAnsh:
## Description

Fixes a bug where CLP pool creation would fail when one of the pair tokens had no existing escrow, and adds comprehensive integration tests covering escrow-validation scenarios for both CLP and CP/Stable pools, now parameterised across chain mode and input token kind (including voucher).

**Root cause:** `ack_concentrated_pool_creation` used `TOKEN_TO_ESCROW.load` (hard failure) instead of `may_load`. When a pool is created with one registered and one new token, the ACK handler would error with `TokenEscrowDoesNotExist` rather than instantiating the missing escrow. CP/Stable's `ack_pool_creation` already handled this correctly.

**Fix:** Mirror `ack_pool_creation` in `ack_concentrated_pool_creation`: use `may_load`, send tokens to the escrow if it exists, otherwise instantiate a new one via `ESCROW_INSTANTIATE_REPLY_ID` with the deposit queued in `PENDING_DEPOSIT_TOKEN`.

**Tests added / updated (CLP):**

| Scenario | Expected | Token kinds | Generated cases |
|---|---|---|---|
| Two fee tiers on same pair | both registered, distinct VLPs | native / smart / voucher × native / smart / voucher | 27 + 27 |
| 1 registered + 1 new token | passes, creates escrow for new token | voucher on registered side, native / smart on new side | 18 |
| Both tokens unregistered | rejected at factory call | native / smart only | 12 |
| Escrow present but denom disallowed | rejected at factory call | native / smart only | 12 |
| Invalid fee-tier / tick-spacing | rejected at factory call | native / smart / voucher × native / smart / voucher | 27 |
| Pool creation error ACK | refunds tokens to sender | native | 1 |

Voucher is excluded from the rejection tests because it changes their failure modes (the router's voucher-existence check fires before the 'two new tokens' rule, and voucher tokens have no escrow for the disallowed check to apply to).

A `TokenEscrowDoesNotExist { token }` error variant was added to surface the token name in diagnostics.

**CP/Stable parity:** `test_create_pool_with_unregistered_token_should_pass_and_create_escrow`, `test_create_pool_with_both_tokens_unregistered_rejected`, and `test_create_pool_with_disallowed_token_rejected` were added across 6 cases each (stable / constant-product × Local / Ibc / Evm).

**Test infrastructure changes** (`tests-integration/src/helpers/factory.rs`, `tests-integration/src/tests_reusable/concentrated_create_pool.rs`):

- `faucet` for `TokenType::Smart` now mints cw20 tokens to the recipient. The previous behaviour set allowance with the sender as its own spender, which cw20 rejects. Minting is the correct semantic match for native's 'add balance' path.
- New `approve_factory_for_smart_tokens` helper grants sender-to-factory allowance for each smart token in a pair, invoked by `create_concentrated_pool_with_tick` and `add_concentrated_liquidity` before the factory is asked to pull those tokens via `TransferFrom`.
- New `prepare_token` / `mint_voucher_balance` helpers wire voucher support: a backing native denom is registered on the factory chain so the router's `TOKEN_DENOMS` has an entry for the token id, and voucher balance is minted on the hub's virtual_balance for the factory sender via a temporary `set_sender` override.
- CLP create-pool tests switched to rstest `#[values]` so that each test expresses its own kind matrix, with `chain_id_for_mode` collapsing the coupled mode / chain_id pair into a single derived argument.
- State loads in `ack_concentrated_pool_creation` are now scoped to the branch that actually needs them.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor/Chore

## Related Issue

N/A

## Testing

```
# Parameterised CLP create-pool suite (123 cases, all pass)
cargo test -p tests-integration concentrated_create_pool

# Escrow scenarios for CP/Stable (all pass)
cargo test -p tests-integration factory_create_pool

# Error ACK refunds tokens to sender (passes)
cargo test -p tests-integration test_ack_error_rolls_back_pending

# Full concentrated and factory swap / add-liquidity suites, no regressions
cargo test -p tests-integration concentrated
cargo test -p tests-integration factory_swap
cargo test -p tests-integration factory_add_liquidity
```

## Checklist

- [x] Self-review completed

### DIFF
--- a/contracts/liquidity/factory/src/ibc/ack_and_timeout.rs
+++ b/contracts/liquidity/factory/src/ibc/ack_and_timeout.rs
@@ -330,8 +330,11 @@ fn ack_concentrated_pool_creation(
                     continue;
                 }
 
-                let escrow_contract =
-                    TOKEN_TO_ESCROW.load(deps.storage, token_info.token.clone())?;
+                let escrow_contract = TOKEN_TO_ESCROW
+                    .load(deps.storage, token_info.token.clone())
+                    .map_err(|_| ContractError::TokenEscrowDoesNotExist {
+                        token: token_info.token.to_string(),
+                    })?;
                 let send_msg = token_info
                     .token_type
                     .create_escrow_msg(token_info.amount, escrow_contract)?;
--- a/contracts/liquidity/factory/src/ibc/ack_and_timeout.rs
+++ b/contracts/liquidity/factory/src/ibc/ack_and_timeout.rs
@@ -324,10 +324,6 @@ fn ack_concentrated_pool_creation(
                 &data.vlp_address,
             )?;
 
-            let state = STATE.load(deps.storage)?;
-            let admins = ADMIN.load(deps.storage)?;
-            let escrow_code_id = state.escrow_code_id;
-
             let mut res = Response::new();
             for token_info in existing_req.pair_info.get_vec_token_info() {
                 if token_info.token_type.is_voucher() {
@@ -345,6 +341,9 @@ fn ack_concentrated_pool_creation(
                         res = res.add_message(send_msg);
                     }
                     None => {
+                        let state = STATE.load(deps.storage)?;
+                        let admins = ADMIN.load(deps.storage)?;
+                        let escrow_code_id = state.escrow_code_id;
                         let init_msg = CosmosMsg::Wasm(WasmMsg::Instantiate {
                             admin: Some(admins.migration_admin.clone().into_string()),
                             code_id: escrow_code_id,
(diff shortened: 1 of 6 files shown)


contracts/liquidity/factory/src/ibc/ack_and_timeout.rs
```diff
@@ -324,18 +324,51 @@ fn ack_concentrated_pool_creation(
                 &data.vlp_address,
             )?;
 
+            let state = STATE.load(deps.storage)?;
+            let admins = ADMIN.load(deps.storage)?;
+            let escrow_code_id = state.escrow_code_id;
```
joemonem: this, admins, and state could be moved to the `None` match branch below

joemonem (approved):

PR #186, opened 2026-04-17
fix: address TODO comments — ack validation, nonpayable guard, stable LP allocation

joemonem:
## Summary

Worked through a series of TODO comments across the codebase, resolving each one with either a code fix, a clarifying comment, or removal where the concern was already addressed.

- **Enforce `nonpayable` on `execute_add_allowed_denom`** (`contracts/liquidity/escrow`): This is a purely administrative function. Sending funds with it would silently lock them in the contract with no recovery path. Added `cw_utils::nonpayable` guard, added `cw-utils` as an escrow dependency, and added `From<PaymentError>` to `ContractError`. Fixed two tests that were inadvertently sending funds during `AddAllowedDenom` calls (via the shared `info` object used for both instantiate and deposit).

- **Remove stale TODO comments**: Removed two TODOs whose concerns were already resolved by existing code — the `ack_swap_request` callback routing (already sends `asset_in` to escrow on success) and the `receive_cw20` zero-amount check (correctly lives at the escrow boundary as a defensive invariant).

- **Use D invariant for LP allocation in stable pools** (`packages/pool`, `contracts/hub/stable_vlp`, `contracts/hub/cp_vlp`): `add_liquidity` was shared between CP and stable VLPs but always used the constant-product formula (geometric mean / proportional min). Stable pools should allocate LP shares proportional to growth in the Curve-style D invariant. Added `calculate_stable_lp_allocation` using `compute_d`, threaded `amp_factor: Option<Uint64>` through `add_liquidity` (mirrors the existing `SwapCalculationMethod` pattern for swaps), and updated the integration test expectation from the incorrect CP value (`31622`) to the correct D-invariant value (`82026`) for a stable pool seeded with 10k/100k tokens at amp=100.

- **Clarify `execute_disallow_denom` behaviour** (`contracts/liquidity/escrow`): Replaced the `//TODO refund the disallowed funds` comment with a note explaining that funds for disallowed denoms remain in escrow until in-flight IBC operations settle, at which point the factory drains them via `execute_withdraw`. Immediate refund on disallow would risk double-spending against pending acks.

## Test plan

- [ ] `cargo test -p router` — ack validation tests pass
- [ ] `cargo test -p escrow` — nonpayable and deposit tests pass
- [ ] `cargo test -p euclid-pool -p cp_vlp -p stable_vlp` — all 79 tests pass
- [ ] `cargo test -p tests-integration` — stable pool integration tests pass with updated LP token expectations
- [ ] `cargo clippy --all-targets --all-features` — no warnings


crnbarr93 (approved):
High level ACK

contracts/hub/router/src/execute/relay.rs
```diff
@@ -204,13 +204,13 @@ pub fn execute_receive_acknowledgement(
...
-    remove_pending_packet_and_decrement_count(deps.storage, &chain_uid, sequence)?;
+    let existing_request =
+        remove_pending_packet_and_decrement_count(deps.storage, &chain_uid, sequence)?;
 
-    // TODO: This is lost during relayer encoding and decoding, fix this once relayer is stable
-    // ensure!(
-    //     existing_request == msg,
-    //     ContractError::new("Ack source msg doesn't match with existing request")
-    // );
+    ensure!(
```
SlayerAnsh: This is not retained properly when evm to cosmos is changed, results in loss of some optional fields resulting in message mismatch. New structure is better but sequence check is better and contracts store the related pending msg in their state to prevent malicious message being passed.

SlayerAnsh (changes requested):

SlayerAnsh (approved):
LGTM!

PR #187, opened 2026-04-17
feat: add missing QueryMsgs for unqueryable state

joemonem:
## Summary

- **Escrow**: `DisallowedDenoms {}`, `GetDenomBalance { denom }`, `GetAllDenomBalances { pagination }` — exposes the disallowed-denom list and per-denom escrow balances (previously completely hidden from clients)
- **Router**: `GetLockedChains {}`, `GetFeeState {}`, `GetDefaultReleaseFee {}`, `GetChainTimeout { chain_uid }` — exposes chain lock status, fee recipients, fallback release fee, and per-chain IBC timeout
- **Factory**: `GetLpShares { vlp }`, `GetRateLimitState {}`, `GetUserRateLimit { user }` — exposes LP share allocation, global rate-limit config, and per-user rate-limit status
- **Orderbook Deposits**: `GetPendingRoot {}`, `GetRootConfig {}` — exposes roots in their challenge window and governance config (permit signer, authorized posters, challenge period)

`GetAllDenomBalances` uses cursor-based `Pagination<String>` (optional) consistent with existing paginated queries. All other new queries are single-item lookups or small admin-controlled lists.

## Test plan

- [ ] `cargo build --all` passes with no errors
- [ ] `cargo test` passes (all 127 router + 61 factory + 43 escrow + other tests green)
- [ ] Verify each new query can be called with `cargo test` or via schema

### DIFF
--- a/contracts/hub/router/src/contract.rs
+++ b/contracts/hub/router/src/contract.rs
@@ -20,6 +20,7 @@ use crate::execute::{execute_manage_router_state, execute_meta_receive, execute_
 
 use crate::query::{
     self, query_all_chains, query_all_escrows, query_all_tokens, query_all_vlps, query_chain,
+    query_chain_timeout, query_default_release_fee, query_fee_state, query_locked_chains,
     query_relayer_addresses, query_release_fees, query_state, query_token_denoms,
     query_token_escrows, query_vlp,
 };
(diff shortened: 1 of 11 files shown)


SlayerAnsh (commented):
Some of the queries were not added to keep the contract size small. For example, LP shares can be calculated via direct state query if needed. Lets analyze what should be included. In general, any query on VSL side should be fine as size restriction is not much concern on VSL but factory side contracts should be as small as possible because same contracts needs to satisfy multiple chain size restrictions.

SlayerAnsh (approved):

PR #189, opened 2026-04-20
Remove factory-side concentrated pool registration check

SlayerAnsh:
## Description

Removes the factory-side check that required a concentrated pool_key to be registered on the local factory before allowing swap hops through it. This check prevented valid swaps that use vouchers or originate from different chains where the pool was never registered locally. Pool existence and validity are still enforced on the VSL/router side.

Also adds a clarifying comment in `concentrated_vlp` about refund behavior with forwarding contracts (refunding to sender can lock vouchers if the forwarding contract does not implement proper voucher handling).

Adds a new router query `GetClpPositionInfo { position_id }` that resolves the concentrated VLP via `CLP_POSITION_ID_VLP_MAP`, cross-queries the VLP for the position, and returns both the VLP address and the position info in a single `ClpPositionInfoResponse`.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor/Chore

## Related Issue

N/A

## Testing

Steps to test:

1. Attempt a concentrated-pool swap hop from a chain where the pool_key is not registered on the local factory (e.g., using vouchers or a cross-chain swap).
2. Verify the swap proceeds and is validated by the router/VLP instead of being rejected at the factory.
3. Query the router with `GetClpPositionInfo { position_id }` for an existing CLP position and confirm the response contains the correct VLP address and position info.
4. Run \`cargo test\` to confirm no regressions.

Automated coverage added in this PR:

- \`factory::execute::swap::tests::test_swap_request_concentrated_pool_key_not_registered_succeeds\` — asserts that \`ExecuteSwapRequest\` with a concentrated \`PoolKey\` absent from \`POOL_KEY_TO_VLP\` is accepted, the pending swap is recorded, and the \`pool_key\` is preserved on the stored \`SwapRequest\`.

## Checklist

- [x] Self-review completed

## Additional Notes

Pool validation still occurs on the hub via the router and the concentrated VLP, so removing the local factory check does not weaken safety; it only removes a false-negative that blocked legitimate cross-chain/voucher flows. The now-unused \`POOL_KEY_TO_VLP\` import in \`contracts/liquidity/factory/src/execute/swap.rs\` is removed as part of this PR.


joemonem:
Shouldn't we add a test case that covers this? Also the PR description mentions a clarifying comment but it isn't present in the code changes

joemonem (approved):

PR #193, opened 2026-04-29
test(clp): add out-of-range position integration tests

SlayerAnsh:
# Pull Request

## Description

Add integration tests for concentrated liquidity positions that are outside the active tick range. Tests cover: active liquidity invariance on add/remove of OOR positions, zero fee accrual for OOR positions, tick crossing activation, and fee collection behavior before and after tick re-entry.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor/Chore

## Related Issue

N/A

## Testing

Steps to test:

1. `cargo test -p tests-integration out_of_range`
2. All 26 test cases should pass (7 test functions × 3 chain modes: Native, IBC, EVM)

## Checklist

- [x] Self-review completed

N/A

## Additional Notes

Tests added in `tests-integration/src/tests_reusable/clp/out_of_range.rs`:

- `test_add_position_below_range_does_not_change_active_liquidity`
- `test_add_position_above_range_does_not_change_active_liquidity`
- `test_out_of_range_position_earns_no_fees`
- `test_swap_into_out_of_range_position_activates_it`
- `test_remove_liquidity_from_out_of_range_position`
- `test_add_to_existing_out_of_range_position_keeps_active_liquidity`
- `test_collect_fees_on_out_of_range_position_returns_zero`
- `test_position_earns_fees_only_after_tick_enters_range`

### DIFF
--- a/tests-integration/src/tests_reusable/clp/authorization.rs
+++ b/tests-integration/src/tests_reusable/clp/authorization.rs
@@ -7,20 +7,14 @@ use euclid::msgs::cross_chain_config::CrossChainConfig;
 use euclid::msgs::factory::msg::QueryMsgFns as FactoryQueryMsgFns;
 use rstest::rstest;
 
+use super::utils::first_position_id;
 use crate::helpers::factory::{
-    add_concentrated_liquidity, create_concentrated_pool, list_position_ids,
-    remove_concentrated_liquidity,
+    add_concentrated_liquidity, create_concentrated_pool, remove_concentrated_liquidity,
 };
 use crate::tests_reusable::concentrated_create_pool::{pair_with_amounts, setup_concentrated_env};
 use crate::tests_reusable::concentrated_swap::execute_concentrated_swap;
 use crate::tests_reusable::constants::{FACTORY_CHAIN_ID_IBC, FACTORY_CHAIN_ID_LOCAL};
 use crate::tests_reusable::factory_register::FactorySetupMode;
-
-fn first_position_id(factory: &factory::FactoryContract<cw_orch::mock::MockBase>) -> Uint128 {
-    let ids = list_position_ids(factory).unwrap();
-    assert!(!ids.is_empty(), "expected at least one position");
-    Uint128::new(ids[0].parse::<u128>().unwrap())
-}
 
 #[cfg(test)]
 mod tests {
--- a/tests-integration/src/tests_reusable/clp/mod.rs
+++ b/tests-integration/src/tests_reusable/clp/mod.rs
@@ -1,6 +1,7 @@
 #![cfg(not(target_arch = "wasm32"))]
 
 pub mod authorization;
+pub mod out_of_range;
 pub mod pending_and_race;
 pub mod position_id;
 pub mod position_lifecycle;
--- a/tests-integration/src/tests_reusable/clp/mod.rs
+++ b/tests-integration/src/tests_reusable/clp/mod.rs
@@ -1,6 +1,7 @@
 #![cfg(not(target_arch = "wasm32"))]
 
 pub mod authorization;
+pub mod one_sided;
 pub mod out_of_range;
 pub mod pending_and_race;
 pub mod position_id;
--- a/tests-integration/src/tests_reusable/clp/one_sided.rs
+++ b/tests-integration/src/tests_reusable/clp/one_sided.rs
@@ -1,0 +1,445 @@
+#![cfg(not(target_arch = "wasm32"))]
+
+use cosmwasm_std::{Addr, Uint128};
+use cw_orch::prelude::*;
+use euclid::msgs::router::query::QueryMsgFns as RouterQueryMsgFns;
+use euclid::msgs::vlp::concentrated::msg::{
+    PositionResponse, QueryMsg as ConcentratedQueryMsg, Slot0Response,
+};
+use rstest::rstest;
+
+use crate::helpers::chains::get_concentrated_vlp;
+use crate::helpers::factory::{
+    add_concentrated_liquidity, create_concentrated_pool, list_position_ids,
+    remove_concentrated_liquidity,
+};
+use crate::tests_reusable::concentrated_create_pool::{pair_with_amounts, setup_concentrated_env};
+use crate::tests_reusable::factory_register::FactorySetupMode;
+
+fn last_position_id(factory: &factory::FactoryContract<cw_orch::mock::MockBase>) -> Uint128 {
+    let ids = list_position_ids(factory).unwrap();
+    assert!(!ids.is_empty(), "expected at least one position");
+    Uint128::new(ids.last().unwrap().parse::<u128>().unwrap())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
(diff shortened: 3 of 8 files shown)


joemonem (commented):
## PR #193 — `test(clp): add out-of-range position integration tests`


### Code Quality

**Inconsistent `rstest` parameterization pattern**

Every other file in `tests_reusable/clp/` (`position_lifecycle.rs`, `authorization.rs`, `position_id.rs`, `pending_and_race.rs`) uses `#[case(mode, factory_chain_id)]` with `factory_chain_id` passed explicitly as a second case parameter:

```rust
// existing pattern
#[rstest]
#[case(FactorySetupMode::Native, FACTORY_CHAIN_ID_LOCAL)]
#[case(FactorySetupMode::Ibc, FACTORY_CHAIN_ID_IBC)]
fn test_foo(#[case] mode: FactorySetupMode, #[case] factory_chain_id: &str) { ... }
```

`out_of_range.rs` uses `#[values(...)]` and calls `mode.factory_chain_id()` internally:

```rust
// new pattern
#[rstest]
fn test_bar(
    #[values(FactorySetupMode::Native, FactorySetupMode::Ibc, FactorySetupMode::Evm)]
    mode: FactorySetupMode,
) {
    let factory_chain_id = mode.factory_chain_id();
    ...
}
```

Both work, but the inconsistency within the same directory makes the codebase harder to navigate. Recommend aligning with the existing pattern.

---

**Duplicated module-level helpers**

`first_position_id` is defined locally in `out_of_range.rs`, `position_lifecycle.rs`, `authorization.rs`, and `concentrated_v3_positions.rs`. `out_of_range.rs` also adds `last_position_id`, `sender`, and `voucher_balance` as new module-level helpers.

These belong in `tests_reusable/clp/utils.rs` (currently only has `price_to_tick` / `pair_to_tick`). The duplication will accumulate over time.

---

**`last_position_id` is order-unsafe**

```rust
fn last_position_id(factory: ...) -> Uint128 {
    let ids = list_position_ids(factory).unwrap();
    Uint128::new(ids.last().unwrap().parse::<u128>().unwrap())
}
```

`list_position_ids` calls `AllTokens` on the CW721-style position token contract. `AllTokens` returns tokens in **lexicographic order**, not insertion order. For IDs "1", "2", ..., "9" this is fine, but once IDs reach "10", lexicographic ordering (`"1", "10", "2", "3", ...`) means `.last()` returns "9", not "10". Should sort numerically before calling `.last()`:

```rust
ids.iter()
    .filter_map(|s| s.parse::<u128>().ok())
    .max()
    .map(Uint128::new)
    .expect("expected at least one position")
```

---

### Correctness & Fragility

**Hardcoded tick-spacing of 10**

Tick alignment is done with:
```rust
let lower = ((slot0.tick - 300) / 10) * 10;
```

This hardcodes `tick_spacing = 10`. If a test pool is ever created with a different spacing, alignment silently breaks. Should align against the pool's actual tick spacing, or at minimum define a named constant `TICK_SPACING` at the top of the file since all pools in this file use `create_concentrated_pool(..., 10, ...)`.

---

**Swap amounts are magic numbers with brittle tick-crossing assumptions**

`test_swap_into_out_of_range_position_activates_it` and `test_position_earns_fees_only_after_tick_enters_range` both rely on a hard-coded `Uint128::new(5_000)` swap being large enough to cross a specific tick boundary (e.g., `upper = slot0.tick - 10`). The post-swap assertion catches a failure — but does so with a `panic!` mid-test rather than surfacing a meaningful "swap was insufficient" error. Consider either using a larger, obviously sufficient swap size, or adding a comment explaining why 5,000 is expected to cross the boundary given 50,000 pool reserves.

---

**Swap return values silently discarded** (6 call sites)

```rust
execute_concentrated_swap(&factory, &router, pool_key.clone(), ...);
// return value dropped
```

If the swap itself errors, the test will continue and produce confusing assertion failures downstream. Consider adding `.unwrap()` (or `.expect("swap should succeed")`) to fail fast.

---

### Testing Coverage

The 8 test scenarios cover the important invariants well:
- OOR below/above doesn't affect active liquidity ✓
- Swaps within range don't accrue fees to OOR positions ✓
- Tick crossing activates the OOR position ✓
- Remove from OOR doesn't affect active liquidity ✓
- Add to existing OOR keeps active liquidity unchanged ✓
- Fee collection on OOR returns zero ✓
- Fee accrual only begins after tick entry ✓

One gap: there's no test for **adding a position that spans the current tick** (in-range), as a positive control for the active-liquidity assertions. The current tests only verify that OOR doesn't change active liquidity, but never confirm a comparable in-range position would.

---

### Minor Issues

- `sender(factory)` is constructed twice per test in `test_collect_fees_on_out_of_range_position_returns_zero` and `test_position_earns_fees_only_after_tick_enters_range` — can be cached as a `let`.

---

### Summary

The test logic is sound and the coverage is meaningful. Main issues to address before merging:

1. **Bug**: `last_position_id` uses lexicographic `.last()` — will give wrong results for IDs ≥ 10.
2. **Consistency**: Use `#[case]` pattern to match all other files in the same directory.
3. **Reliability**: Call `.unwrap()` on `execute_concentrated_swap` so test failures are clear.
4. **Maintainability**: Extract `first_position_id`, `last_position_id`, `sender`, `voucher_balance` to `utils.rs`.

SlayerAnsh:
## Review Response

### Addressed

**1. `last_position_id` lexicographic bug** — Fixed. Now uses `.filter_map(|s| s.parse::<u128>().ok()).max()` for numeric ordering. Also caught same bug in `one_sided.rs` and an inline variant in `position_lifecycle.rs`.

**4. Extract helpers to `utils.rs`** — Done. Moved `first_position_id`, `last_position_id`, `sender`, `voucher_balance` to `utils.rs`. Removed duplicates from 6 files.

**5. Hardcoded tick spacing** — Added `const TICK_SPACING: i64 = 10;` in `out_of_range.rs`, replaced all tick alignment math.

Commit: `9aa49348`

### Not addressing

**2. `#[case]` vs `#[values]` pattern** — `one_sided.rs` also uses `#[values]` + `mode.factory_chain_id()`. This is the newer pattern in this directory, not an inconsistency. It's less boilerplate (no redundant `factory_chain_id` parameter, derived from mode). Older files can be migrated later.

**3. Swap return values discarded** — `execute_concentrated_swap` returns `Uint128`, not `Result`. It panics internally on failure via `.unwrap()` calls. There's nothing to `.unwrap()` on the caller side. `position_lifecycle.rs` and `pending_and_race.rs` also discard the return value.

**6. Magic swap amounts** — 5,000 is 10% of 50,000 pool reserves. The assertions (`assert!(slot0_after.tick < upper, ...)`) already catch insufficient swaps with descriptive messages. Adding a comment explaining the math would be more noise than signal.

**7. `sender()` called twice** — Trivial mock state reads. No observable cost.

**8. In-range positive control** — In-range position behavior is already covered by `position_lifecycle.rs` (add liquidity, fee accrual, collect fees). Adding it here would duplicate existing coverage.

joemonem (approved):

PR #196, opened 2026-05-19
fix(euclid): reorg-safe tx_id format with per-sender nonce

joemonem:
## Description

Makes `tx_id` deterministic under reorg replay (SC-8, CosmWasm slice). The current `generate_tx` format embeds `block.height` and `transaction.index`, both of which can shift across reorg replays, leaving in-flight cross-chain operations stuck when the source chain reorganises after the destination has already ack'd. This PR removes the block-derived fields and changes the nonce from a global counter to a per-sender map, so the same logical transaction reproduces the same `tx_id` under any realistic reorg.

**Format change:** `{sender}:{chain_id}:{height}:{index}:{nonce}` → `{sender}:{chain_id}:{nonce}`

**State change:** `TX_NONCE: Item<u128>` → `TX_NONCES: Map<String, u128>` keyed by `sender.to_sender_string()`. The old global Item at storage key `"tx_nonce"` is left orphaned (zero reads/writes); the new Map uses namespace `"tx_nonces"`.

Old in-flight `PENDING_*` entries written under the previous 5-field format remain valid (segment-count differs from new 3-field format, no key collision). No `MigrateMsg` payload, no chain pause, no `PENDING_*` rewriting.

Solana and Tron are out of scope for this PR — handled separately under SC-8 parent.

## Type of Change

- [x] Bug fix
- [x] Breaking change (event/wire format for ``tx_id`` — opaque end-to-end inside contracts, but off-chain consumers must handle both 3- and 5-segment shapes during transition)

## Related Issue

- Linear: SC-8 (Block Reorg Fix — remove block height from txid format)
- Slice breakdown and rationale in ``TX_ID_ADJUSTMENT_ISSUES.md`` at repo root

### Unit tests (``packages/euclid/src/utils/tx.rs``)
- ``id_equality_under_input_variation`` rstest matrix: determinism under varying ``block.height``/``transaction.index``; sender disambiguation; chain-id disambiguation.
- ``nonce_disambiguates_back_to_back_calls``: same-Env calls produce distinct ids (nonce regression guard).
- ``output_format_has_three_top_level_segments``: format-shape pin.
- ``per_sender_nonce_is_independent_across_senders``: Alice/Bob interleaving — both at nonces 1, 2.
- ``per_sender_nonce_stable_under_cross_sender_reordering``: replaying Bob-then-Alice doesn't shift Alice's first-call id.

### Synthetic stuck-packet tests (router unit level)
- ``test_swap_reply_ok_missing_pending_returns_error``: ack for unknown ``tx_id`` on ``PENDING_SWAPS`` fails predictably, no silent mutation, seeded entry untouched.
- ``test_release_ack_with_unknown_tx_id_fails_cleanly``: same for ``PENDING_RELEASE_VOUCHER``.

### Layer 1 cw-orch-interchain tests (``tests-integration/src/tests_reusable/tx_id_format.rs``)
- ``tx_id`` emitted in outbound IBC packet has the new 4-segment shape (expanded).
- Third segment is ``chain_id``, not block height — field-ordering guard.
- ``TX_NONCES`` storage populated per-sender after a real cross-chain call.
- Sequential calls increment per-sender counter 1→2 end-to-end through ack delivery.
- Old global ``TX_NONCE`` key never written by new code (orphan-storage invariant).

### Suggested follow-on testing (operational, outside this PR)
- **Layer 2 — devnet upgrade rehearsal:** deploy old code, leave 2–3 packets un-acked, run ``MigrateMsg``, deliver held-back acks, assert they resolve. The only test that exercises the upgrade boundary itself.
- **Layer 3 — testnet shadow:** snapshot ``PENDING_*`` + old ``TX_NONCE`` before upgrade, monitor segment-count histogram and ack-handler error rate for 24h post-upgrade.
- **Off-chain consumer coordination** (relayer, indexer, dashboards) before any testnet deploy — they need to handle both 3- and 5-segment ``tx_id`` shapes during transition.

## Pre-PR Checklist

- [x] ``cargo test`` passes (euclid: 119, router: 100, factory: 98, tests-integration: 5 new + 1776 existing)
- [ ] ``cargo test --features full_decimals`` passes — not run locally; please confirm in CI
- [x] ``cargo fmt --check`` passes
- [ ] ``cargo test -p tests-fuzz`` passes — not run locally; please confirm in CI

## Checklist

- [x] Self-review completed
- [x] Tests added/updated
- [x] Documentation updated (inline doc on ``generate_tx``; ``TX_ID_ADJUSTMENT_ISSUES.md`` at root)
- [x] ``CHANGELOG.md`` updated under ``[euclid]`` in ``### Fixed``

## Additional Notes

**Deploy invariant:** the format change (slice 1) and per-sender nonce (slice 3) must ship together. Shipping slice 1 alone would create a window where post-upgrade new-format tx_ids could collide with pre-upgrade in-flight ones (both 3-field with global nonce); both commits are on this branch to avoid that.

**Rollback story:** ``MigrateMsg`` is a no-op for these changes, so rollback is "just deploy old code." But any new-format packets in flight at the moment of rollback would become stuck under the old version (old code looks up old-format keys; new-format acks miss). Layer 2 devnet rehearsal is the last opportunity to catch problems before commit-to-forward.


SlayerAnsh (commented):
## Reorg Scenario Coverage Review

The tx_id determinism properties are well-tested (unit + integration). What's missing is testing the **dedup rejection paths** that actually fire during a reorg, and verifying the **end-to-end behavioral guarantees** the fix is supposed to provide.

### Cross-chain operation timeline for reference

```
Factory chain              Relayer              Hub (Router)
─────────────────────────────────────────────────────────────
[1] User tx executes
    → generate tx_id
    → store PENDING_SWAPS
    → emit send_packet
                           [2] Relay →
                                                [3] Router receives
                                                    → sequence dedup check
                                                    → tx_id dedup check
                                                    → store PENDING_SWAPS
                                                    → VLP logic
                                                    → emit ack
                           [4] Relay ack ←
[5] Factory receives ack
    → load PENDING_SWAPS
    → remove PENDING_SWAPS
    → release/refund
```

---

### Scenario A — Pre-relay reorg (between [1] and [2])

**Same tx replayed:** Factory regenerates same tx_id (nonce rolled back too). Relayer may forward events from both old and new blocks. Router's `CROSS_CHAIN_PROCESSED_RECEIVED_PACKETS[(chain_uid, sequence)]` catches the duplicate. Ack from the one Router did process finds the correct `PENDING_SWAPS` entry on Factory (rebuilt with same key). **Safe at contract level.**

**Tx dropped from new block:** Old packet event still visible to relayer. Router processes it, sends ack. Factory has no `PENDING_SWAPS` entry → ack handler `.load()` fails. **Funds stuck on Router side.** (Operational, not fixable at contract level; relayer must wait for finality.)

**Missing tests:**
- [ ] Router rejects a second packet with same `(chain_uid, sequence)` through `execute_receive_packet` — integration test that delivers a duplicate sequence and asserts `"Processed sequence already exists"` error.
- [ ] Factory ack handler errors cleanly when `PENDING_SWAPS` entry is absent (simulates "tx dropped" case) — currently only Router side has `test_swap_reply_ok_missing_pending_returns_error`.

---

### Scenario B — Mid-processing reorg (between [2] and [4], the primary reorg scenario)

**Same tx replayed:** Router already stored `PENDING_SWAPS[tx_id]` from original packet. Replayed packet from new block rejected by sequence dedup at `router/execute/relay.rs:116-121`. Ack returns → Factory has rebuilt `PENDING_SWAPS[(sender, tx_id)]` → processes correctly.

This is the core scenario the fix targets. The properties hold, but no test exercises them end-to-end.

**Missing tests:**
- [ ] Full Scenario B flow: Factory sends swap → relay to Router → Router stores `PENDING_SWAPS[tx_id]` → attempt to deliver same packet again (same sequence) → assert `"Processed sequence already exists"` → deliver ack back to Factory → assert Factory processes ack and removes `PENDING_SWAPS`.
- [ ] Router `TxAlreadyExist` rejection: Deliver two swap packets with the **same `tx_id` but different sequences** to Router. First succeeds, second fails with `TxAlreadyExist`. This guards against a relayer that somehow assigns a new sequence to the replayed packet. (`router/ibc/receive/swap.rs:48-51` has the guard, zero tests exercise it.)

---

### Scenario C — Post-complete reorg (full round-trip done, then Factory reorgs)

**Same tx replayed:** Router already removed `PENDING_SWAPS[tx_id]` during ack processing, already released output tokens/vouchers. Factory state rolled back → input tokens un-escrowed. Replayed packet gets rejected by Router sequence dedup, but the first processing already released funds. **Cross-chain state inconsistency (user gets free tokens).**

This is fundamentally a finality problem, not solvable at contract level. But it's worth documenting as a known limitation and verifying the dedup still fires (Router doesn't accidentally re-process).

**Block height concern:** `CROSS_CHAIN_PROCESSED_RECEIVED_PACKETS` stores `Uint256::from(env.block.height)` as its value, but the dedup check is `.has()` (existence only). Different block heights between old and new factory blocks don't affect Router-side dedup. **No breakage from height difference.**

**Missing tests:**
- [ ] Assert that after a full round-trip (packet + ack processed), the Router still rejects a re-delivery of the same `(chain_uid, sequence)`. Confirms dedup survives past ack removal of `PENDING_SWAPS`.
- [ ] Document Scenario C as a known limitation in the PR or in code comments. Mitigation is purely operational (relayer finality window).

---

### Scenario D — Ack re-delivery after Factory reorg

Factory processed ack → `CROSS_CHAIN_PROCESSED_RECEIVED_PACKETS` on Factory recorded the sequence. Reorg rolls this back. Same ack could be re-delivered.

If same tx replayed: `PENDING_SWAPS` exists again → ack processes normally → **correct**.
If tx dropped: `PENDING_SWAPS` missing → ack `.load()` fails → **error, but no double-spend**.

**Missing tests:**
- [ ] Factory rejects a duplicate inbound ack sequence — integration test that delivers the same ack sequence twice to Factory and asserts `"Processed sequence already exists"`.

---

### Scenario E — Timeout race with reorg

Factory sends packet with timeout. Reorg happens. Same tx replays with same timeout (timeout is hub-chain-relative, not factory-relative). Two packets in flight with same sequence. Router dedup catches duplicate.

Edge case: if original packet timed out on Router side, `PENDING_SWAPS` was never stored. Replayed packet with same sequence still rejected by sequence dedup (the timeout handler still marks the sequence as processed).

**Missing test:**
- [ ] Verify that after a timeout fires for a given sequence, re-delivery of the same `(chain_uid, sequence)` is still rejected. (Timeout handler may or may not write to `CROSS_CHAIN_PROCESSED_RECEIVED_PACKETS`; needs verification.)

---

### Scenario F — `TxAlreadyExist` guards across ALL operation types

The `TxAlreadyExist` guard exists in multiple Factory execute paths, all untested for the rejection case:

| Path | Location |
|------|----------|
| `execute_swap` | `factory/execute/swap.rs:104-106` |
| `execute_deposit_token` (3 paths) | `factory/execute/token.rs:52, 159, 247` |
| `execute_create_pool` | `factory/execute/pool.rs:150` |
| `execute_add_liquidity` | `factory/execute/pool.rs:250` |
| `execute_remove_liquidity` | `factory/execute/pool.rs:384` |
| Router `receive_swap` | `router/ibc/receive/swap.rs:48-51` |

**Missing test:**
- [ ] At minimum, one integration test per operation type that attempts to submit a duplicate `tx_id` and asserts `TxAlreadyExist`. These guards are the application-layer safety net when sequence dedup somehow doesn't fire.

---

### Summary

The **determinism properties** (same inputs → same tx_id) are solid. What's untested is every **dedup rejection path** that actually fires during a reorg:

1. Router sequence dedup (`CROSS_CHAIN_PROCESSED_RECEIVED_PACKETS`) — no integration test
2. Router `TxAlreadyExist` on swap receive — zero tests
3. Factory `TxAlreadyExist` on all operation types — zero tests
4. Factory inbound ack sequence dedup — no integration test
5. Post-complete reorg dedup survival — no test
6. End-to-end "send → reject duplicate → ack still works" flow — no test

**Recommendation:** Scenarios B and C are highest priority since they're the exact reorg patterns this fix targets. The cw-orch mock can't truly reorg, but we can simulate by manually constructing and delivering duplicate packets. The unit test level (`MockDeps`) is sufficient for most of these.

SlayerAnsh (approved):
## Verified

### Core fix correctness
- `generate_tx` produces `{sender}:{chain_id}:{nonce}`, no block-derived components. Format pinned by `output_format_has_three_top_level_segments` unit test.
- `TX_NONCES: Map<String, u128>` keyed per-sender. Old `TX_NONCE: Item<u128>` at `"tx_nonce"` orphaned, never read/written by new code (integration test pins this via raw storage query).
- Per-sender nonce isolation confirmed: cross-sender reordering does not shift any individual sender's nonce stream.

### Migration safety
- `tx_id` is fully opaque throughout the codebase. Zero production code parses, splits, or inspects its content. Every usage is map key lookup, event attribute, or struct field pass-through.
- Old 5-segment in-flight PENDING_* entries coexist safely with new 4-segment entries (different segment count, no key collision possible).
- Ack handlers load by exact key from IBC packet payload, which carries the original tx_id verbatim. Old in-flight packets resolve correctly post-upgrade.
- No `MigrateMsg`, no chain pause, no PENDING_* rewriting required.

### Dedup rejection paths (Scenarios A through F)
- **A (pre-relay reorg):** Router sequence dedup via `CROSS_CHAIN_PROCESSED_RECEIVED_PACKETS` rejects duplicate `(chain_uid, sequence)`. Factory ack errors cleanly when PENDING row missing post-reorg.
- **B (mid-processing reorg):** Full flow tested: duplicate rejected, original ack still resolves. Router `TxAlreadyExist` guard on `PENDING_SWAPS.has(tx_id)` tested for same-tx_id-different-sequence case.
- **C (post-complete reorg):** Sequence dedup survives past ack removal of PENDING entry. Known limitation (finality problem) documented in `generate_tx` doc comment.
- **D (ack re-delivery):** Factory rejects duplicate ack after round-trip completes.
- **E (timeout race):** Sequence dedup survives after timeout fires.
- **F (TxAlreadyExist across all ops):** All 7 factory execute paths covered (swap, register_denom, deregister_denom, deposit_token, request_pool_creation, add_liquidity, remove_liquidity) plus Router swap receive.

### Test results
- `cargo test -p euclid -- tx`: 7/7 passed
- `cargo test -p router` (dedup tests): 3/3 passed
- `cargo test -p factory -- duplicate_tx_id`: 7/7 passed
- `cargo test -p tests-integration -- tx_id_format reorg_dedup`: 11/11 passed

### Minor notes (non-blocking)
- `predicted_tx_id` in `factory_ack_errors_cleanly_when_pending_missing_post_reorg` works because chain_uid == chain_id in test setup. If they ever diverge, `Map::remove` silently becomes a no-op. Worth a comment or extracting tx_id from events instead.
- `wrapping_add` rationale is documented in `TX_ID_ADJUSTMENT_ISSUES.md` out-of-scope section, which is sufficient.

